PR #21, opened 2021-03-11
Add invoke update

JonZeolla:
# Contributor Comments
Add `invoke update` (used via `pipenv run invoke update`) to maintain the project dependencies.

This reverts the version pinning of packages added in #23 due to maintenance concerns.

## Pull Request Checklist
Thank you for submitting a contribution to `goat`!

In order to streamline the review of your contribution we ask that you review and comply with the below requirements:
- [X] Rebase your branch against the latest commit of the target branch, which likely should be `main`
- [X] If there is an issue associated with your Pull Request, align your PR title with [this documentation](https://help.github.com/en/articles/closing-issues-using-keywords)

PR #22, opened 2021-03-15
Remove versioning

JonZeolla:
# Contributor Comments
Remove versioning from the project.  Depends on #21 for security updates

## Pull Request Checklist
Thank you for submitting a contribution to `goat`!

- [X] Rebase your branch against the latest commit of the target branch, which likely should be `main`
- [X] If there is an issue associated with your Pull Request, align your PR title with [this documentation](https://help.github.com/en/articles/closing-issues-using-keywords)


tasks.py
```diff
@@ -146,16 +144,26 @@ def run_security_tests(*, image: str):
 
 # git
 REPO = git.Repo(CWD)
-COMMIT_HASH = REPO.head.object.hexsha
+COMMIT_HASH = REPO.head.object.hexsha[:7]
```
ericSeiso: Add support for short hash instead? Unsure if it matter for this size of repo.

JonZeolla: This is changing the hash into a "short hash" and the number of characters is the git default (although `git` is smart enough to pick a longer hash if it detects a collision, and this is not).

Let me make it smarter

JonZeolla: It is now smarter and uses [a dynamic config](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreabbrev) by default

ericSeiso (approved):
LGTM

PR #55, opened 2021-11-11
Fix the goat

JonZeolla:
# Contributor Comments

Fix CI failures in the goat

## Pull Request Checklist

Thank you for submitting a contribution to the goat!

In order to streamline the review of your contribution we ask that you review and comply with the below requirements:

- [X] Rebase your branch against the latest commit of the target branch
- [X] If you are adding a dependency, please explain how it was chosen
- [X] If manual testing is needed in order to validate the changes, provide a testing plan and the expected results
- [X] If there is an issue associated with your Pull Request, align your PR
  title with [this documentation](https://help.github.com/en/articles/closing-issues-using-keywords)

PR #57, opened 2022-03-11
Added exceptions to permit apt-get

callmestu:
# Contributor Comments

[Please insert a description of the change here]

## Pull Request Checklist

- [x] If there is an issue associated with your Pull Request, align your PR
  title with [this documentation](https://help.github.com/en/articles/closing-issues-using-keywords)

### DIFF
--- a/etc/.dockerfilelintrc
+++ b/etc/.dockerfilelintrc
@@ -2,3 +2,8 @@
 rules:
   # Workaround until replicatedhq/dockerfilelint#169 is resolved
   missing_tag: off
+  # Two apt-get exemptions below are to prioritize component updates/maintenance over hermetically sealed, reproducable builds
+  # Permit apt-get update to be used generally
+  apt-get-update_require_install: off
+  # Permit apt-get upgrade to apply patches to runner node
+  apt-get-upgrade: off
--- a/etc/.dockerfilelintrc
+++ b/etc/.dockerfilelintrc
@@ -2,5 +2,5 @@
 rules:
   # Workaround until replicatedhq/dockerfilelint#169 is resolved
   missing_tag: off
-  # Permit apt-get upgrade to apply patches to prioritize component updates/maintenance over hermetically sealed, reproducable builds
+  # Permit apt-get upgrade to apply patches to prioritize component updates/maintenance over reproducable builds
   apt-get-upgrade: off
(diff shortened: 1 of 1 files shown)


etc/.dockerfilelintrc
```diff
@@ -2,3 +2,8 @@
 rules:
   # Workaround until replicatedhq/dockerfilelint#169 is resolved
   missing_tag: off
+  # Two apt-get exemptions below are to prioritize component updates/maintenance over hermetically sealed, reproducable builds
+  # Permit apt-get update to be used generally
+  apt-get-update_require_install: off
```
JonZeolla: Apt update just updates the local package index for later installs. I don't know of any situation where we would want to do an update but not as install. Also, since we delete the package index at the end of a RUN, a new update would be required for later installs, meaning this rule is always valuable to leave to the default on, as far as I can tell 

Was there a use case you were thinking of that I'm not?

JonZeolla (changes requested):

etc/.dockerfilelintrc
```diff
@@ -2,3 +2,8 @@
 rules:
   # Workaround until replicatedhq/dockerfilelint#169 is resolved
   missing_tag: off
+  # Two apt-get exemptions below are to prioritize component updates/maintenance over hermetically sealed, reproducable builds
```
JonZeolla: Hermetic builds are separate from the exemptions; just running an apt-get install from a remote destination would fail a hermetic build.  In those cases you download and validate, then pass the files explicitly into the build process to be hermetic

```suggestion
  # The apt-get upgrade exemption below is to prioritize component updates/maintenance over reproducable builds
```

etc/.dockerfilelintrc
callmestu: This is resolved - we added a package with `apt-get install`

etc/.dockerfilelintrc
callmestu: Language updated.

JonZeolla: Sounds good.  For the future, you can literally commit the patch that I suggested above directly to your branch without needing to even leave the browser

JonZeolla:
@callmestu this is now blocked on https://github.com/SeisoLLC/goat/pull/55

Feel free to update Wrike accordingly.  I believe PR 55 has it's own task you can set as a dependency

CC @seanctech

JonZeolla (approved):

PR #65, opened 2022-07-21
added isort config file

seanctech:
# Contributor Comments

Added isort config file to set profile to be compatible with black

## Pull Request Checklist

### DIFF
--- a/etc/.isort.cfg
+++ b/etc/.isort.cfg
@@ -1,0 +1,2 @@
+[tool.isort]
+profile = "black"


etc/.isort.cfg
```diff
@@ -0,0 +1,2 @@
+[tool.isort]
```
JonZeolla, line 1: Should we align more closely with https://raw.githubusercontent.com/github/super-linter/main/TEMPLATES/.isort.cfg ?

JonZeolla: I think the answer is no, that the profile=black is simply a better approach to what they have upstream in super-linter

seanctech: Yeah, isort specifically has a profile for black compatibility that they maintain. It's probably easier to let them handle the compatibility and just leverage their work for the profile.

JonZeolla (approved):

PR #70, opened 2022-08-17
Ignore

digeseiso:
# Contributor Comments

Added ignore for workflows folder

## Pull Request Checklist

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -32,8 +32,8 @@ jobs:
       - name: Install the dependencies
         run: |
           python -m pip install --upgrade pipenv
-          pipenv uninstall setuptools
-          pipenv install setuptools==58.1.0
+          python -m pip uninstall setuptools
+          python -m pip  setuptools==58.1.0
           pipenv install --python ${{ env.python_version }} --deploy --ignore-pipfile --dev
       - name: Build the goat
         run: pipenv run invoke build
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -62,8 +62,6 @@ jobs:
             ${{ runner.os }}-python-
       - name: Install dependencies
         run: |
-          python -m pip uninstall -y setuptools
-          python -m pip install setuptools==65.1.0
           python -m pip install --upgrade pipenv
           pipenv install --python ${{ env.python_version }} --deploy --ignore-pipfile --dev
       - name: Build the goat
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -10,7 +10,7 @@ on:
       - main
 
 env:
-  python_version: "3.10"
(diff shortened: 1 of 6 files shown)


JonZeolla:
The issue is early enough in the pipeline that I think we just need to figure out how other pipelines are installing deps with pipenv without hitting this bug

.gitignore
```diff
@@ -1,3 +1,4 @@
+.DS_Store
```
JonZeolla: Shouldn't we just update the rest of this to be the content from https://www.toptal.com/developers/gitignore/api/vim,emacs,vs,python,node,macos ?

Note that removes docker because `#!! ERROR: docker is undefined. Use list command to see defined gitignore types !!#`

etc/.jscpd.json
```diff
@@ -0,0 +1,10 @@
+{
+    "threshold": 0,
+    "reporters": [
+        "consoleFull"
+    ],
+    "ignore": [
+        "**/workflows/**"
```
JonZeolla: Very generic, why wouldn't we say .github/workflows/** ?

.github/workflows/commit.yml
```diff
@@ -62,6 +62,8 @@ jobs:
             ${{ runner.os }}-python-
       - name: Install dependencies
         run: |
+          python -m pip uninstall -y setuptools
+          python -m pip install setuptools==58.1.0
```
JonZeolla: Do we have an issue to point to in setuptools or cpython regarding this issue?

JonZeolla (changes requested):

.gitignore
JonZeolla: Still a thing

etc/.jscpd.json
```diff
@@ -0,0 +1,10 @@
+{
+    "threshold": 0,
+    "reporters": [
+        "consoleFull"
+    ],
+    "ignore": [
+        ".github/workflows/**"
```
JonZeolla: How do we know this works?

digeseiso: I built it and tested it locally

JonZeolla: 🤘

.github/workflows/commit.yml
JonZeolla: https://github.com/pypa/setuptools/issues/3278

JonZeolla:
The backlog task to clean up this patch version pin is https://www.wrike.com/open.htm?id=950047944

JonZeolla (dismissed):
Don't forget to change your commit from ignore to something more meaningful when squash merging

etc/.jscpd.json
digeseiso: @JonZeolla re-opening this because the suggestion didnt work, i have modified it to reflect whats working

JonZeolla (approved):

PR #86, opened 2023-02-18
feat: Migrate to taskfiles

seanctech:
# Contributor Comments

Added centralized Taskfiles to be used by dependent repos generated from cookiecutter.

## Pull Request Checklist


JonZeolla:
`Unknown word (Taskfiles)`: Gotta update the dictionary

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - >
+        docker run --rm -v ${PWD}:/goat -e RUN_LOCAL=true
+        --entrypoint black seiso/goat:latest .
+
+  update:
+    cmds:
+      - >
+        docker run  --rm -v ${PWD}:/usr/src/app -w /usr/src/app python:{{.PYTHON_VERSION}} /bin/bash
+        -c "python3 -m pip install --upgrade pipenv &>/dev/null && pipenv update"
+
```
JonZeolla:
```suggestion
```

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,53 @@
...
+  VERSION: null
+  PYTHON_VERSION: null
+
+tasks:
+  lint:
+    preconditions:
+      - sh: test `git ls-files -z -o --exclude-standard | wc -w` = "0"
+        msg: "There are untracked files, please commit before linting."
+    cmds:
+      - docker run --rm -v ${PWD}:/goat -e RUN_LOCAL=true -e EXCLUDE="/goat" seiso/goat:latest
```
JonZeolla:
```suggestion
      - docker run --rm -v "${PWD}:/goat" -e RUN_LOCAL=true -e EXCLUDE="/goat" seiso/goat:latest
```

JonZeolla: What is the `-e EXCLUDE="/goat"`?

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+vars:
+  IMAGE_NAME: null
+  VERSION: null
+  PYTHON_VERSION: null
+
+tasks:
+  lint:
+    preconditions:
+      - sh: test `git ls-files -z -o --exclude-standard | wc -w` = "0"
```
JonZeolla: What's the benefit of this over:

> test `git status -s | wc -w` = "0"

?

It just seems like it would be simpler/easier to understand by the reader.

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,53 @@
...
+        msg: "There are untracked files, please commit before linting."
+    cmds:
+      - docker run --rm -v ${PWD}:/goat -e RUN_LOCAL=true -e EXCLUDE="/goat" seiso/goat:latest
+
+  test:
+    preconditions:
+      - sh: test `git symbolic-ref -q HEAD`
+        msg: "In detached HEAD state."
+    cmds:
+      - pipenv run pytest tests 2>&1 >/dev/null
```
JonZeolla: Shouldn't we allow the test output to be shown in the logs for troubleshooting?
```suggestion
      - pipenv run pytest tests
```

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+  build:
+    cmds:
+      - docker build -t {{.IMAGE_NAME}}:latest -t {{.IMAGE_NAME}}:{{.VERSION}} .
```
JonZeolla: If we're going to be able to use `docker` again, we should enable buildkit.  It's often on by default, but some of those details are out of our control, so we can do something like:
```suggestion
      - DOCKER_BUILDKIT=1 docker build -t {{.IMAGE_NAME}}:latest -t {{.IMAGE_NAME}}:{{.VERSION}} .
```
Or something like:
```suggestion
      - docker buildx build -t {{.IMAGE_NAME}}:latest -t {{.IMAGE_NAME}}:{{.VERSION}} .
```

We should weigh the pros and cons of each and pick the best for our situation.

Also, this needs to identify if this is a release or not, and set a variable accordingly for the buildargs.  Right now, VERSION and COMMIT_HASH must be passed in as buildargs with the right information.  Details [here](https://github.com/SeisoLLC/cookiecutter-python/blob/085215423d3c66a42408cc7d2a0b0eb18b62c0f6/%7B%7Bcookiecutter.project_name%7Creplace(%22%20%22%2C%20%22%22)%7D%7D/tasks.py#L114).

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+  reformat:
+    cmds:
+      - >
+        docker run --rm -v ${PWD}:/goat -e RUN_LOCAL=true
```
JonZeolla:
```suggestion
        docker run --rm -v "${PWD}:/goat" -e RUN_LOCAL=true
```

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,53 @@
...
+        --entrypoint isort seiso/goat:latest .
+        --settings-file /action/lib/.automation/.isort.cfg
+      - >
+        docker run --rm -v ${PWD}:/goat -e RUN_LOCAL=true
```
JonZeolla:
```suggestion
        docker run --rm -v "${PWD}:/goat" -e RUN_LOCAL=true
```

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,53 @@
...
+        --entrypoint black seiso/goat:latest .
+
+  update:
+    cmds:
+      - >
+        docker run  --rm -v ${PWD}:/usr/src/app -w /usr/src/app python:{{.PYTHON_VERSION}} /bin/bash
```
JonZeolla:
```suggestion
        docker run  --rm -v "${PWD}:/usr/src/app" -w /usr/src/app python:{{.PYTHON_VERSION}} /bin/bash
```

JonZeolla (changes requested):
Some quick thoughts

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,53 @@
+# https://taskfile.dev
+
+version: '3'
+
+vars:
+  IMAGE_NAME: null
+  VERSION: null
+  PYTHON_VERSION: null
+
+tasks:
```
JonZeolla: This is missing the `release` and `publish` tasks

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,53 @@
+# https://taskfile.dev
```
JonZeolla: Is it possible to do another level of includes here?  I wonder if a base `clean` is possible in `goat/Task/Taskfile.yaml`, which this uses and then adds on the python-specific items?  Then we could move thumbs, ds_store, and super-linter.log to that, and when we move away from `super-linter` the super-linter.log would only need to be cleaned up from one place (and any new general cleanups would just be added there as well).

Task/python/Taskfile.yaml
seanctech: I agree here, that should be the final version. The current usage is really the end result of trying a few different approaches that were a bit more explicit but they didn't work as well as I would have liked.

Task/python/Taskfile.yaml
seanctech: I've added a base clean, currently it's leveraging python to clean because globstar isn't consistent across all shells. Will need to find a better way to deal with this but for now leveraging I think leveraging the repo python is okay.

JonZeolla: @seanctech I don't understand; if we can run a docker container then we control the shell and globstar config?

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+includes:
+  base: 
+    taskfile: ../Taskfile.yaml
+    internal: true
+
+vars:
+  IMAGE_NAME: null
+  PYTHON_VERSION: null
+  PROJEC_SLUG: null
```
JonZeolla: Typo?
```suggestion
  PROJECT_SLUG: null
```

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,134 @@
...
+        msg: "There are untracked files, please commit before linting."
+    cmds:
+      - docker run --rm -v "${PWD}:/goat" -e RUN_LOCAL=true seiso/goat:latest
+
+  test:
+    preconditions:
+      - sh: test `git symbolic-ref -q HEAD`
+        msg: "In detached HEAD state."
+    cmds:
+      - pipenv run pytest tests
```
JonZeolla: Don't we need to add `--cov={{.PROJECT_SLUG}}`?

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,134 @@
...
+          {{range $tag := .REPO_TAGS | splitLines -}}
+          repo_tags.append("{{$tag}}")
+          {{end}}
+          if (version_string in repo_tags and "{{.TAG_COMMIT_HASH}}" == "{{.COMMIT_HASH}}"):
+              build_version = "{{.VERSION}}"
+          else:
+              build_version = f"{{.VERSION}}-{{.COMMIT_HASH_SHORT}}"
+          print(build_version)'
+    cmds:
+      - docker buildx build --tag {{.IMAGE_NAME}}:latest --tag {{.IMAGE_NAME}}:{{.BUILD_VERSION}} .
```
JonZeolla: This assumes there is only one `BUILD_VERSION`, which is fine for now.  In the future we'll need to support something more dynamic

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+  reformat:
+    cmds:
```
JonZeolla: Not sure if this syntax is right.  I assume if the `sh` fails, then the msg is printed and it doesn't move to the `cmds`.
```suggestion
    preconditions:
      - sh: docker pull seiso/goat:latest
        msg: "Unable to pull the latest version of the goat."
    cmds:
```

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,134 @@
...
+  PYTHON_VERSION: null
+  PROJEC_SLUG: null
+  REPO_TAGS:
+    sh: git tag -l
+
+tasks:
+  lint:
+    preconditions:
+      - sh: test `git status -s | wc -w` = "0"
+        msg: "There are untracked files, please commit before linting."
```
JonZeolla:
```suggestion
        msg: "There are untracked files, please commit before linting."
      - sh: docker pull seiso/goat:latest
        msg: "Unable to pull the latest version of the goat."
```

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,134 @@
...
+    preconditions:
+      - sh: test `git status -s | wc -w` = "0"
+        msg: "There are untracked files, please commit before creating a release."
+    vars:
+      RELEASE_TYPE: '{{.CLI_ARGS}}'
+    cmds:
+      - pipenv run bumpversion --config-file setup.cfg {{.RELEASE_TYPE}}
+
+  publish:
+    deps: [build]
```
JonZeolla: This is definitely in the realm of opinion, but I prefer to be more explicit about `task build publish` instead of `task publish` and it does a `task build` for me.  I assume you disagree?  I could be convinced

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,134 @@
...
+        docker run --rm -v "${PWD}:/goat" -e RUN_LOCAL=true
+        --entrypoint black seiso/goat:latest .
+
+  update:
+    cmds:
+      - >
+        docker run  --rm -v "${PWD}:/usr/src/app" -w /usr/src/app python:{{.PYTHON_VERSION}} /bin/bash
+        -c "python3 -m pip install --upgrade pipenv &>/dev/null && pipenv update"
+
+  release-calver:
```
JonZeolla: Is there a way to set a variable input of calver vs semver, make this just `release`, and have some logic below that decides on calver vs semver?

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,134 @@
...
+    preconditions:
+      - sh: test `git status -s | wc -w` = "0"
+        msg: "There are untracked files, please commit before creating a release."
+    vars:
+      DATE_INFO: '{{now | date "2006.01"}}'
```
JonZeolla: That's an interesting syntax...  Is this YYYY.MM?

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,134 @@
...
+      RELEASE_VERSION:
+        sh: |
+          docker run --rm -v "${PWD}:/goat" python:{{.PYTHON_VERSION}}-slim python -c '
```
JonZeolla: If we're volume mounting into `/goat` and then running a `python` image, the working dir will be different.  We need to specify a `-w` for the working dir to align them, and I recommend we don't use `/goat` in non-goat images.

JonZeolla (changes requested):
This was a line-by-line review.  I still need to take a step back and think more critically.  I think this is going great and in the right direction though

Task/python/Taskfile.yaml
JonZeolla: Or:
```suggestion
    preconditions:
      - sh: docker pull seiso/goat:latest || echo "Failed to pull the latest goat, continuing anyway..."
    cmds:
```

Task/python/Taskfile.yaml
JonZeolla: Or:
```suggestion
        msg: "There are untracked files, please commit before linting."
      - sh: docker pull seiso/goat:latest || echo "Failed to pull the latest goat, continuing anyway..."
```

Task/python/Taskfile.yaml
JonZeolla:
```suggestion
      - pipenv run pytest --cov={{.PROJECT_SLUG}} tests
```

Task/python/Taskfile.yaml
JonZeolla:
```suggestion
```

Task/python/Taskfile.yaml
JonZeolla:
```suggestion
          docker run --rm -v "${PWD}:/usr/src/app" -w /usr/src/app python:{{.PYTHON_VERSION}}-slim python -c '
```

Task/python/Taskfile.yaml
JonZeolla: We should be able to pass a valid `fmt` from [here](https://github.com/go-task/slim-sprig/blob/52ccab3ef572c7e1a2c258be183f9a9296d60152/date.go#L14)

Task/python/Taskfile.yaml
JonZeolla: Let's write up some logic that parses the `setup.cfg` per https://github.com/SeisoLLC/cookiecutter-python/pull/79#issuecomment-1471999665

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,122 @@
...
+  REPO_TAGS:
+    sh: git tag -l
+  VERSIONING:
+    sh: grep versioning setup.cfg | cut -d '"' -f 2
+
+tasks:
+  lint:
+    preconditions:
+      - sh: test `git status -s | wc -w` = "0"
+        msg: "There are untracked files, please commit before linting."
```
digeseiso:
```suggestion
```

digeseiso: @seanctech this wont work because the command returns an error code 1 upon `echo $?` because we are in detached head state in CI when it runs.

seanctech: @digeseiso just needs to be changed, it should be successful on 1. Detached HEAD would be 0. I think I intended to write not equal to 0 but did the opposite, I'll get it fixed.

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,120 @@
...
+        docker run --rm -v "${PWD}:/goat" -e RUN_LOCAL=true
+        --entrypoint isort seiso/goat:latest .
+        --settings-file /action/lib/.automation/.isort.cfg
+      - >
+        docker run --rm -v "${PWD}:/goat" -e RUN_LOCAL=true
+        --entrypoint black seiso/goat:latest .
+
+  update:
+    cmds:
+      - >
```
JonZeolla:
```suggestion
      - git submodule update --init --recursive
      - >
```

Task/Taskfile.yaml
```diff
@@ -0,0 +1,10 @@
+# https://taskfile.dev
```
JonZeolla: Can we rename to Taskfile.yml?  I will hate my life forever if we don't nip this in the bud

Task/python/Taskfile.yaml
```diff
@@ -0,0 +1,121 @@
+# https://taskfile.dev
```
JonZeolla: Can we rename to Taskfile.yml?  I will hate my life forever if we don't nip this in the bud

JonZeolla (changes requested):

Task/Taskfile.yaml
seanctech: Sure thing, when you init a new Taskfile it names it "Taskfile.yaml" but I checked the style guide and it has "Taskfile.yml".

Task/python/Taskfile.yml
```diff
@@ -0,0 +1,121 @@
...
+          {{range $tag := .REPO_TAGS | splitLines -}}
+          repo_tags.append("{{$tag}}")
+          {{end}}
+          if (version_string in repo_tags and "{{.TAG_COMMIT_HASH}}" == "{{.COMMIT_HASH}}"):
+              build_version = "{{.VERSION}}"
+          else:
+              build_version = f"{{.VERSION}}-{{.COMMIT_HASH_SHORT}}"
+          print(build_version)'
+    cmds:
+      - docker buildx build --tag {{.IMAGE_NAME}}:latest --tag {{.IMAGE_NAME}}:{{.BUILD_VERSION}} .
```
JonZeolla: Our standard `Dockerfile`s expect us to pass in a `VERSION` and `COMMIT_HASH` as build args.  See https://github.com/SeisoLLC/cookiecutter-python/blob/fe92d43a70ccb648f2701d622246c388db5f39d5/%7B%7Bcookiecutter.project_name%7Creplace(%22%20%22%2C%20%22%22)%7D%7D/Dockerfile#L13-L17

JonZeolla (changes requested):

Task/python/Taskfile.yml
```diff
@@ -0,0 +1,121 @@
...
+        --settings-file /action/lib/.automation/.isort.cfg
+      - >
+        docker run --rm -v "${PWD}:/goat" -e RUN_LOCAL=true
+        --entrypoint black seiso/goat:latest .
+
+  update:
+    cmds:
+      - git submodule update --init --recursive
+      - >
+        docker run  --rm -v "${PWD}:/usr/src/app" -w /usr/src/app python:{{.PYTHON_VERSION}} /bin/bash
```
JonZeolla:
```suggestion
        docker run --rm -v "${PWD}:/usr/src/app" -w /usr/src/app python:{{.PYTHON_VERSION}} /bin/bash
```

JonZeolla (changes requested):

Task/python/Taskfile.yml
JonZeolla:
```suggestion
      - docker buildx build --build-arg VERSION="{{.BUILD_VERSION}}" --build-arg COMMIT_HASH="{{.COMMIT_HASH}}" --tag {{.IMAGE_NAME}}:latest --tag {{.IMAGE_NAME}}:{{.BUILD_VERSION}} .
```

JonZeolla (approved):

PR #90, opened 2023-03-25
feat: build oci images, not docker image format

JonZeolla:
# Contributor Comments

If you don't specify to build in the oci format, you build docker image formats, which are mostly interoperable, but we prefer oci images.

## Pull Request Checklist

### DIFF
--- a/Task/python/Taskfile.yml
+++ b/Task/python/Taskfile.yml
@@ -54,7 +54,7 @@ tasks:
               build_version = f"{{.VERSION}}-{{.COMMIT_HASH_SHORT}}"
           print(build_version)'
     cmds:
-      - docker buildx build --build-arg VERSION="{{.BUILD_VERSION}}" --build-arg COMMIT_HASH="{{.COMMIT_HASH}}" --tag {{.IMAGE_NAME}}:latest --tag {{.IMAGE_NAME}}:{{.BUILD_VERSION}} .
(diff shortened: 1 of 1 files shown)


Task/python/Taskfile.yml
```diff
@@ -54,7 +54,7 @@ tasks:
               build_version = f"{{.VERSION}}-{{.COMMIT_HASH_SHORT}}"
           print(build_version)'
     cmds:
-      - docker buildx build --build-arg VERSION="{{.BUILD_VERSION}}" --build-arg COMMIT_HASH="{{.COMMIT_HASH}}" --tag {{.IMAGE_NAME}}:latest --tag {{.IMAGE_NAME}}:{{.BUILD_VERSION}} .
+      - docker buildx build -o type=oci --build-arg VERSION="{{.BUILD_VERSION}}" --build-arg COMMIT_HASH="{{.COMMIT_HASH}}" --tag {{.IMAGE_NAME}}:latest --tag {{.IMAGE_NAME}}:{{.BUILD_VERSION}} .
```
seanctech, line 57: Doesn't the OCI option require a destination tarball for the image?

https://docs.docker.com/build/exporters/oci-docker/

JonZeolla: Wow; this is going to be really non-optimal.

First of all, you can't load it into the docker daemon directly.  Like you mention, it needs to go to a file, like:

```bash
docker buildx build -o type=oci,dest=example.tar .
```

But after that, you can't even `docker load` it without losing some of its `oci`-ness<sup>TM</sup>

So you need a second tool like `crane` to push it.  Hard pass; way too complicated.  Closing the PR

PR #87, opened 2023-02-24
adding ruby for linting ruby files

digeseiso:
# Contributor Comments

This PR updates the super-linter to the latest version, installs ruby and the following gems:

- Robocop
- Rubocop-github

These installations will allow the Goat to lint ruby files, see here for failure/use case:

https://github.com/SeisoLLC/inspec/actions/runs/4256809121/jobs/7406156043#step:4:1115

## Pull Request Checklist

Thank you for submitting a contribution to the goat!

In order to streamline the review of your contribution we ask that you review and comply with the below requirements:

- [x] Rebase your branch against the latest commit of the target branch
- [x] If you are adding a dependency, please explain how it was chosen
- [x] If manual testing is needed in order to validate the changes, provide a testing plan and the expected results
- [x] If there is an issue associated with your Pull Request, link the issue to the PR.

PR #94, opened 2023-05-01
feat: add support for dynamic list of linters to replace the super-linter

ChiefHolland:
# Contributor Comments

The changes included in this PR refactor the current goat to support a dynamic list of linters and their respective arguments with the intent of replacing the super-linter. In the current goat build, the super-linter runs a series of linters sequentially, followed by a seiso_lint function that adds some hard-coded linting. One of the problems with the current setup is when a file fails to pass linting, the goat exits and has to be re-run to catch the next failure. Another problem with the current goat is its speed. Running the commands sequentially is slow, and if the process has to re-run multiple times, it becomes a burden on developers.  

The solution presented to each of these problems is as follows:  
1. Exiting on failures: Log exit codes from each step of the linting process and do not exit the goat until the exit codes are examined for non-zero at the end of the process.  
2. Speed: Create a file with a delimited list of linter and any necessary metadata for desired output and load that list dynamically as an associative array. This, along with specific logic in the application flow, allows extensibility in the goat and the ability to run linters in parallel, increasing the processing speed.  

TODO: Incorporate the dynamic list of linters into the Dockerfile installation method to allow for a single point of modification for added/removed linters.  

This PR is the first stage in moving away from the super-linter and relates to the following issues:  
1. <https://github.com/github/super-linter/issues/3477#issuecomment-1317763372>
2. <https://github.com/SeisoLLC/goat/pull/87#issuecomment-1472955208>
3. <https://github.com/SeisoLLC/goat/pull/87#issuecomment-1476909616>
4. <https://github.com/SeisoLLC/goat/commit/65a72d8d647bd7254b820dbf1386ca20cd4b2965>

## Pull Request Checklist


.pre-commit-config.yaml
ChiefHolland, line 1: I added some pre-commit functionality to help during the contributor process. This does not include a complete list of possible linters with pre-commit options, but was included as a PoC to see how the addition could improve development time.

.vscode/extensions.json
ChiefHolland, line 1: This was auto corrected by a linter on pre-commit

Dockerfile
```diff
@@ -23,19 +23,20 @@ ENV PIP_NO_CACHE_DIR=1
...
-                       npm \
-                       tini \
-                       ruby \
- && gem install \
-                       rubocop \
-                       rubocop-github \
- && npm install --no-cache -g dockerfile_lint \
-                       cspell \
-                       markdown-link-check
+    && pipenv install --deploy --ignore-pipfile \
```
ChiefHolland, line 26: These formatting changes were auto-corrected on pre-commit. I'm commenting on these to increase visibility for the pre-commit addition to see if it is viable and what tweaks might be necessary in the future.

Dockerfile
```diff
@@ -23,19 +23,20 @@ ENV PIP_NO_CACHE_DIR=1
...
+    && pip install shed ruff \
```
ChiefHolland, line 27: Added packages for new linters.

Task/Taskfile.yml
```diff
@@ -1,6 +1,7 @@
+---
 # https://taskfile.dev
 
-version: '3'
+version: "3"
```
ChiefHolland, line 4: This was a pre-commit auto-fix

Task/python/Taskfile.yml
```diff
@@ -1,6 +1,7 @@
+---
 # https://taskfile.dev
 
-version: '3'
+version: "3"
```
ChiefHolland, line 4: This was a pre-commit auto-fix

Task/python/Taskfile.yml
```diff
@@ -54,7 +55,8 @@ tasks:
               build_version = f"{{.VERSION}}-{{.COMMIT_HASH_SHORT}}"
           print(build_version)'
     cmds:
-      - docker buildx build --build-arg VERSION="{{.BUILD_VERSION}}" --build-arg COMMIT_HASH="{{.COMMIT_HASH}}" --tag {{.IMAGE_NAME}}:latest --tag {{.IMAGE_NAME}}:{{.BUILD_VERSION}} .
+      - docker buildx build --build-arg VERSION="{{.BUILD_VERSION}}" --build-arg COMMIT_HASH="{{.COMMIT_HASH}}" --tag \
```
ChiefHolland, line 58: Line length problem was discovered on pre-commit, so this was a manual change to correct

Task/python/Taskfile.yml
```diff
@@ -80,7 +82,7 @@ tasks:
       - sh: test `git status -s | wc -w` = "0"
         msg: "There are untracked files, please commit before creating a release."
     vars:
-      RELEASE_TYPE: '{{.CLI_ARGS}}'
+      RELEASE_TYPE: "{{.CLI_ARGS}}"
```
ChiefHolland, line 85: Pre-commit auto-fix

Task/python/Taskfile.yml
```diff
@@ -104,7 +106,8 @@ tasks:
           new_version = f"{{.DATE_INFO}}.{increment}"
           print(new_version)'
     cmds:
-      - '{{if eq .VERSIONING "CalVer"}}pipenv run bumpversion --config-file setup.cfg --new-version {{.RELEASE_VERSION}} unusedpart{{else if eq .VERSIONING "SemVer"}}pipenv run bumpversion --config-file setup.cfg {{.RELEASE_TYPE}}{{end}}'
+      - '{{if eq .VERSIONING "CalVer"}}pipenv run bumpversion --config-file setup.cfg --new-version {{.RELEASE_VERSION}} \
```
ChiefHolland, line 109: Also a manual change to fix a pre-commit error

entrypoint.sh
```diff
@@ -18,7 +18,6 @@ function feedback() {
   case "${1}" in
     ERROR)
       >&2 echo -e "${!color}${1}:  ${2}${DEFAULT}"
-      exit 1
```
ChiefHolland, line 21: This exit was causing the goat to exit on the first error found. I removed it in order to handle exiting an reporting failures downstream.

etc/.yamllint.yml
ChiefHolland, line 1: This was included to match up some line length consistency across Seiso. I want to fix this across all repos and all linters to ensure consistency and document line length in our developer documentation. Is 160 an acceptable threshold for all filetypes?

etc/links.json
```diff
@@ -1,11 +1,17 @@
...
-    {"pattern": "^https://github.com/SeisoLLC/"},
-    {"pattern": "^https://raw.githubusercontent.com/SeisoLLC/"},
-    {"pattern": "^https://seiso.sharepoint.com/"}
-  ],
-  "timeout": "20s",
-  "retryOn429": true,
-  "retryCount": 3,
-  "fallbackRetryDelay": "5s"
-}
+    "ignorePatterns": [
```
ChiefHolland, line 2: This formatting change was a pre-commit auto-fix

package.json
ChiefHolland, line 1: This was an auto-fix during pre-commit

entrypoint.sh
```diff
@@ -100,37 +99,131 @@ function super_lint() {
   /action/lib/linter.sh
 }
 
+function get_files_matching_filetype() {
```
ChiefHolland, line 124: This function compiles a list of files to be linted based on filetype metadata passed in from the linters.txt file.

entrypoint.sh
```diff
@@ -100,37 +99,131 @@ function super_lint() {
...
+    filename=$(basename "$file")
+
+    if [[ "$filename" == *"$filetype" ]]; then
+      matching_files+=("$file")
+    fi
+  done
+  echo "${matching_files[@]}"
+}
+
+function linter_failed() {
```
ChiefHolland: This function takes the exit code from each linter process and adds the linter and the code to an array to be processed when all of the linting has completed.

entrypoint.sh
```diff
@@ -100,37 +99,131 @@ function super_lint() {
...
+
+  if [[ "${return:-1}" != 0 ]]; then
+    linter_exit_codes+=(["${linter[name]}"]="${return}")
+  else
+    return "${return}"
+  fi
+}
+
+function lint_files() {
+  exec &>> "${linter[logfile]}"
```
ChiefHolland: This function takes arguments from the linter array, dynamically builds the command, evaluates the command, and passes the exit code to the linter_failed function for logging. It takes into account three conditions: 1) a linter is internally handling all file types to determine which ones to lint indicated by 'filetype=all' in linters.txt; 2) a linter is being run on specific files with, what I'm calling, an executor (i.e. npx--I am completely open to changing this terminology); and 3) the linter is being run on specific files as a standalone command (no executor).

entrypoint.sh
```diff
@@ -100,37 +99,131 @@ function super_lint() {
...
       continue
     fi
-
     included+=("${file}")
+  done < <(find . -path "./.git" -prune -or -type f)
+  
+  declare -gA linter_exit_codes
+  input="/etc/opt/goat/linters.txt"
 
-    # Check Dockerfiles
```
ChiefHolland, line 116: Replacing this call with the dynamic one

entrypoint.sh
```diff
@@ -100,37 +99,131 @@ function super_lint() {
...
-    # Check .md file spelling and links
-    if [[ "${file}" = *.md ]]; then
-      npx cspell -c /etc/opt/goat/cspell.config.js -- "${file}"
-      npx markdown-link-check --config /etc/opt/goat/links.json --verbose "${file}"
+    IFS="," read -ra pairs <<< "$line"
+ 
+    unset linter
+    declare -gA linter
+
+    for pair in "${pairs[@]}"; do
```
ChiefHolland: Clear the linter array for each line in linters.txt and read in the values for the current run

entrypoint.sh
```diff
@@ -100,37 +99,131 @@ function super_lint() {
...
+      IFS="=" read -r key value <<< "$pair"
+      linter["$key"]="$value"
+    done
+    
+    linter+=([logfile]="/etc/opt/goat/logs/${linter[name]}.log")
+    echo "===============================" >> "${linter[logfile]}"
+    echo "Running linter: ${linter[name]}"
+    echo "${linter[name]^^}" >>"${linter[logfile]}"
+    
+    lint_files &
```
ChiefHolland: Background each linter pass and capture the PID to check error code, then loop through the exit code array to see which ones failed.

entrypoint.sh
```diff
@@ -100,37 +99,131 @@ function super_lint() {
...
+    lint_files & 
+    
+    echo "-------------------------------" >> "${linter[logfile]}"
+  done < $input
+
+  wait
+
+  cat /etc/opt/goat/logs/*
+  
+  echo -e "\nScanned ${#included[@]} files"
```
ChiefHolland, line 192: Some of the original seiso_lint function was retained in this refactor

entrypoint.sh
```diff
@@ -100,37 +99,126 @@ function super_lint() {
...
   echo "Excluded ${#excluded[@]} files"
+
+  for lint in "${!linter_exit_codes[@]}"; do
```
ChiefHolland: The remainder of the script handles smoother exiting. Exit codes provide feedback per linter feedback on failures for the seiso_lint, and keeping the super-linter from exiting allows us to control exiting for either linter set.

etc/linters.txt
```diff
@@ -0,0 +1,5 @@
+#name=shed,args=--refactor,filetype=all,installer=pip
```
ChiefHolland: This is a PoC for how we can add linters. The installer field is set to eventually making the installation dynamic in the Dockerfile.

tasks.py
```diff
@@ -27,7 +27,7 @@ def opinionated_docker_run(
     auto_remove: bool = False,
     detach: bool = True,
     environment: dict = {},
-    entrypoint: str = None,
+    entrypoint: str = "",
```
ChiefHolland, line 30: This was a manual change of a pre-commit finding.

Task/python/Taskfile.yml
```diff
@@ -104,7 +106,8 @@ tasks:
...
+        unusedpart{{else if eq .VERSIONING "SemVer"}}pipenv run bumpversion --config-file setup.cfg {{.RELEASE_TYPE}}{{end}}'
 
   publish:
     cmds:
```
JonZeolla, line 113: We need to update the `clean` task below to cleanup the ruff cache also.

Dockerfile
JonZeolla: LGTM

etc/linters.txt
JonZeolla: I like the idea but let's avoid txt files.  Can we do something more structured without adding any more runtime dependencies, like json, yml, toml, cfg, ini, etc.?

tasks.py
JonZeolla: Eww, I don't like empty string in this case.  But I get why they do it. Can you point me to the specific tool and finding ID?

entrypoint.sh
```diff
@@ -100,37 +99,126 @@ function super_lint() {
   /action/lib/linter.sh
 }
 
+function get_files_matching_filetype() {
+  local matching_files=()
+  local filenames=("${@:2}")
+  local filetype=$1
```
JonZeolla: I think a better way to do this and the prior line would be to replace it with this so we don't need to slice the array:

```suggestion
  local filetype=$1
  shift
  local filenames=("$@")
```

entrypoint.sh
```diff
@@ -100,37 +99,126 @@ function super_lint() {
...
+
+  for file in "${filenames[@]}"; do
+    filename=$(basename "$file")
+
+    if [[ "$filename" == *"$filetype" ]]; then
```
JonZeolla, line 133: nit: This reason that this works is because the `*` is a glob pattern, but using `=~` is better because it uses regex and we can also be explicit about the end of file with $, although there is no functional issue with what's currently there.

```suggestion
    if [[ "$filename" =~ .*${filetype}$ ]]; then
```

JonZeolla (changes requested):
Partial review

entrypoint.sh
JonZeolla: Actually, let's stick with the globbing approach otherwise we'll need to escape all of the `.`s that you have in `filetype`

entrypoint.sh
```diff
@@ -100,37 +99,128 @@ function super_lint() {
   /action/lib/linter.sh
 }
 
+function get_files_matching_filetype() {
+  local matching_files=()
+  local filenames=("${@:2}")
```
```suggestion
```

tasks.py
ChiefHolland:
```
2023-05-02 18:40:15 [ERROR]   Found errors in [mypy] linter!  
2023-05-02 18:40:15 [ERROR]   Error code: 1. Command output:  
------
tasks.py:30: error: Incompatible default for argument "entrypoint" (default has type "None", argument has type "str")  [assignment]  
tasks.py:30: note: PEP 484 prohibits implicit Optional. Accordingly, mypy has changed its default to no_implicit_optional=True  
tasks.py:30: note: Use https://github.com/hauntsaninja/no_implicit_optional to automatically upgrade your codebase
```

JonZeolla: Ew.  Ok.  I expected `docker-py` to type this as `None | str` but I guess not.

Task/Taskfile.yml
```diff
@@ -1,10 +1,12 @@
...
 
 tasks:
   clean:
     cmds:
       - find . -type f -name '.DS_Store' -delete
       - find . -type f -name '.Thumbs.db' -delete
       - find . -type f -name 'super-linter.log' -delete
+      - find . -type d -name '.ruff_cache' -exec rm -r "{}" \;
```
ChiefHolland, line 12: Added a cleanup task for .ruff_cache. Another option is to run ruff with --no-cache.

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
     fi
-  done < <(find . -path "./.git" -prune -or -type f)
-
-  echo "Scanned ${#included[@]} files"
-  echo "Excluded ${#excluded[@]} files"
+    
+    unset linter
+    declare -gA linter
+ 
+    while IFS='=' read -r key value; do
```
ChiefHolland, line 179: Refactored to process linters from json instead of txt.

etc/linters.txt
ChiefHolland: Done

Task/python/Taskfile.yml
ChiefHolland: Added a line to remove .ruff_cache

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
+    lint_files & 
+    pid=$!
+    pids["$pid"]+="${linter[name]}"
+
+    echo "-------------------------------" >> "${linter[logfile]}"
+  done < <(jq -c '.[]' $input)
+
+  for p in "${!pids[@]}"; do
+    set +e
+    wait "$p"
```
JonZeolla, line 199: If the exit code is 0 isn't there where we should print the associated output?  Then we also have control over ordering, instead of doing the `cat /etc/opt/goat/logs/*` below.

Also, logs should go into `/opt/` maybe `/opt/goat/log/goat.log`?

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
 }
 
+linter_failed="false"
 
 setup_environment
 check_environment
-super_lint
+
+set +e
+super_linter_result=$(super_lint >/dev/null; echo $?)
```
JonZeolla: Wouldn't this `/dev/null` the super linter logs?  I suggest we send stdout and stderr to a file and cat it below (don't forget both!)

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
+    
+    if [[ "$filename" == *"$filetype" ]]; then
+      matching_files+=("$file")
+    fi
+  done
+  echo "${matching_files[@]}"
+}
+
+function linter_failed() {
+  return=$1
```
JonZeolla: Still need to quote the inputs

Task/Taskfile.yml
JonZeolla: Let's let caches stay unless we clean 'em up

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
   /action/lib/linter.sh
 }
 
+function get_files_matching_filetype() {
+  local filetype=$1
+  shift
+
+  matching_files=()
+
+  for file in "$@"; do
```
JonZeolla: I would make this a variable and then loop through the variable just for readability

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
   /action/lib/linter.sh
 }
 
+function get_files_matching_filetype() {
+  local filetype=$1
```
JonZeolla: Please double quote

entrypoint.sh
JonZeolla: Also we should keep these as locals

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
+      matching_files+=("$file")
+    fi
+  done
+  echo "${matching_files[@]}"
+}
+
+function linter_failed() {
+  return=$1
+  shift 
+  name=$1
```
JonZeolla: For something simple like this we can keep the shift out and just:
```suggestion
  name="$2"
```

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
+    if [[ "$filename" == *"$filetype" ]]; then
+      matching_files+=("$file")
+    fi
+  done
+  echo "${matching_files[@]}"
+}
+
+function linter_failed() {
+  return=$1
+  shift
```
```suggestion
```

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
+    set -e
+  done
+
+  cat /etc/opt/goat/logs/*
+  
+  echo -e "\nScanned ${#included[@]} files"
+  echo -e "Excluded ${#excluded[@]} files\n"
 }
 
+linter_failed="false"
```
JonZeolla: We're now overloading a variable and function name, let's make them different.  I think the var name makes more sense and the function should adjust

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
+    filename=$(basename "$file")
+    
+    if [[ "$filename" == *"$filetype" ]]; then
+      matching_files+=("$file")
+    fi
+  done
+  echo "${matching_files[@]}"
+}
+
+function linter_failed() {
```
JonZeolla: We should not overload vars and function names.  Really this is something like:
```suggestion
function check_linter_exit_code() {
```

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
+  return=$1
+  shift 
+  name=$1
+
+  if [[ "${return:-1}" != 0 ]]; then
+    linter_exit_codes+=(["$name"]="${return}")
+  else
+    return "${return}"
```
JonZeolla: Hmm, what are we doing here? Maybe we move the `linter_failed` variable setting logic to here?

JonZeolla: But also, why do we not return anything explicitly if the exit code was non-zero?

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
+
+for lint in "${!linter_exit_codes[@]}"; do
+  if [[ "${linter_exit_codes[${lint}]}" -gt 0 ]]; then
+    feedback_label="ERROR"
+    linter_failed="true"
+  else
+    feedback_label="DEBUGGING"
+  fi
+
+  feedback "${feedback_label}" "${lint} discovered errors."
```
JonZeolla: Hmm, not true, right?  This gets called when feedback_label is DEBUGGING, which only happens if the linter exit code is not greater than 0, meaning it didn't discover errors

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
+done
+
+if [ "${linter_failed:-false}" == "true" ]; then
```
JonZeolla: We should always default closed, meaning if linter_failed was for whatever reason not set, it would fail the pipeline.

etc/linters.json
```diff
@@ -0,0 +1,28 @@
+[
```
JonZeolla, line 1: SO much more readable, thank you :)

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
+        
+    linter+=([logfile]="/etc/opt/goat/logs/${linter[name]}.log")
+
+    echo "===============================" >> "${linter[logfile]}"
+    echo "Running linter: ${linter[name]}"
+    echo "${linter[name]^^}" >>"${linter[logfile]}"
+
+    lint_files & 
+    pid=$!
+    pids["$pid"]+="${linter[name]}"
```
JonZeolla: Why are we `+=`ing here? From what I can see, `pids` is a global associative array, and we're appending a string to an empty string which is the default value. We can just do a `=` assignment.
```bash
jonzeolla@~$ echo ${thearray[c]}

jonzeolla@~$ thearray[c]+='example'
jonzeolla@~$ echo ${thearray[c]}
example
```

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
-      dockerfile_lint -f "${file}" -r /etc/opt/goat/oci.yml
-    fi
+  input="/etc/opt/goat/linters.json"
 
-    # Check .md file spelling and links
-    if [[ "${file}" = *.md ]]; then
-      npx cspell -c /etc/opt/goat/cspell.config.js -- "${file}"
-      npx markdown-link-check --config /etc/opt/goat/links.json --verbose "${file}"
+  while read -r line; do
+    if [[ $line == \#* ]]; then
```
JonZeolla: I think this is a remnant of the old `txt` approach, right?  Since the JSON is multi-line, commenting things out in that way would be confusing.

JonZeolla: If we want to allow disabling, I would be more explicit by checking for a value in the JSON which says something like `"enabled": "true"`.  That said, we probably don't need this and can just delete elements from the JSON list when we no longer need them.

entrypoint.sh
JonZeolla: Executor is fine.

I'm not a huge fan of the `exec` approach on this first line though, changing the stdout and stderr of the current runtime entirely to the log file for the provided linter.  If anything else ever happens in between as a part of the goat, it would get sucked into that log file.

Instead I recommend we just redirect the `eval`s

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
+    return "${return}"
+  fi
+}
+
+function lint_files() {
+  exec &>> "${linter[logfile]}"
+  if [[ "${linter[filetype]}" = "all" ]]; then
+    cmd="${linter[name]} ${linter[args]}"
+    eval "$cmd"
+  else
```
JonZeolla: Instead of this else we could just return, then `fi`, and have the following `for` loop one level out-dented for readability.  <3 guard clauses

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
+    for file in $(get_files_matching_filetype "${linter[filetype]}" "${included[@]}"); do 
+      if [[ "${linter[executor]+x}" ]]; then
```
JonZeolla: We should document somewhere (not sure where?) in a md that executors are optional

JonZeolla (changes requested):
This is a 90% review.  After feedback is addressed I'll wrap up the review and test as well.  Thanks

JonZeolla:
@ChiefHolland don't forget to resolve your merge conflicts as well.

entrypoint.sh
ChiefHolland: Just curious, if we write to the same file in parallel tasks, would we potentially run into a problem with conflicting processes trying to write at the same time? This was a problem for Windows development, so I opted to write to individual files and throw all of the output out at the end.

entrypoint.sh
ChiefHolland: I was thinking we could do with a CONTRIBUTING.md for this project.

entrypoint.sh
ChiefHolland: I added the log output here and did away with the check_linter_exit_code function by moving the logic here. There is a near duplication with the super-linter check, but that will be removed in the next phase.

JonZeolla: @ChiefHolland it would still be a different log per linter.  The `goat.log` at the end was a mistake because I got distracted mid-review 🤐

entrypoint.sh
JonZeolla: @ChiefHolland Agreed

.pre-commit-config.yaml
```diff
@@ -0,0 +1,53 @@
+minimum_pre_commit_version: "2.9.0"
+repos:
+  - repo: https://github.com/ansible/ansible-lint
+    rev: v6.14.4
```
JonZeolla, line 4: Do we have a way to maintain these?  Since this is always in repos[].rev and repos[].repo points to a github repo it should be pretty simple by reusing things we've already done, if done in python.

entrypoint.sh
```diff
@@ -18,7 +18,6 @@ function feedback() {
   case "${1}" in
     ERROR)
       >&2 echo -e "${!color}${1}:  ${2}${DEFAULT}"
-      exit 1
       ;;
     WARNING)
       >&2 echo -e "${!color}${1}:  ${2}${DEFAULT}"
```
JonZeolla, line 23: The `SUPPRESS_POSSUM` below should move into the `super_linter` function to simplify future removal

JonZeolla: Same with the `cp -p` line (which I will be SO glad to remove once the super-linter is gone)

JonZeolla: Possibly the `GITHUB_WORKSPACE` item as well?  I forget

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
+  echo -e "\nScanned ${#included[@]} files"
+  echo -e "Excluded ${#excluded[@]} files\n"
 }
 
-
 setup_environment
 check_environment
-super_lint
+
+superlinter_logfile="/opt/goat/log/super-linter.log"
```
JonZeolla: All the superlinter specific logic should go to the `super_lint` function to segment it

JonZeolla: So this should pretty much end up looking like:
```bash
...
check_environment
superlinter_result=$(super_lint ...)
if [ "${superlitner_result" -gt 0 ]; then
...
fi
seiso_lint
```

Meaning that the logfile management, set -e, and set +e, etc. happens inside of super_lint.

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
+    if [[ "${linter[executor]+x}" ]]; then
+      cmd="${linter[executor]} ${linter[name]} ${linter[args]} ${file}"
+    else
+      cmd="${linter[name]} ${linter[args]} ${file}"
+    fi
+    eval "$cmd" >> "${linter[logfile]}" 2>&1
+  done
+}
+
+linter_failed="false"
```
JonZeolla: This should get moved to the setup_environment section. As long as that is run before `super_lint`, `seiso_lint`, etc. we should be all set.

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
+  input="/etc/opt/goat/linters.json"
+
+  while read -r line; do  
+    unset linter
+    declare -gA linter
+ 
+    while IFS='=' read -r key value; do
+      value=$(echo "$value" | tr -d "'")
+      linter["$key"]=$value
+    done < <(echo "$line" | jq -r 'to_entries|map("\(.key)=\(.value|tostring)")|.[]')
```
JonZeolla, line 182: You are now a `jq` ninja.

entrypoint.sh
```diff
@@ -100,37 +99,139 @@ function super_lint() {
...
+        
+    linter+=([logfile]="/opt/goat/log/${linter[name]}.log")
```
JonZeolla: 👀 I never knew about this format - interesting.

Why not use the standard `linter[logfile]="/opt/goat/log/${linter[name]}.log"`?

.pre-commit-config.yaml
JonZeolla: Also, wouldn't it just be easier to run the `goat` via pre-commit so we know it's aligned with what will happen in the pipelines?  Then we will just have one entry here with all the components bundled, configured, etc.

JonZeolla (changes requested):

etc/linters.json
JonZeolla, line 1: We should order this so the faster ones are first - it will give us a (very minor) performance improvement.  I think:
- ruff
- dockerfile_lint
- markdown-link-check
- cspell

entrypoint.sh
ChiefHolland: I will add this in the next round of approvals.

.pre-commit-config.yaml
ChiefHolland: This is the dream. I think the `goat` would be perfect as a pre-commit! As to the maintenance of pre-commit repos, I'm not sure. I'm game to build that into a later sprint, though.

entrypoint.sh
ChiefHolland: Done

entrypoint.sh
ChiefHolland: I was making it a part of the associative array, but you're right. It isn't necessary and will be reassigned a value each loop.

etc/linters.json
ChiefHolland: Done

entrypoint.sh

entrypoint.sh

entrypoint.sh

.pre-commit-config.yaml
ChiefHolland: To be moved to CC...

entrypoint.sh

entrypoint.sh
ChiefHolland: Refactored and removed this function

entrypoint.sh
ChiefHolland: Removed

entrypoint.sh
ChiefHolland: Refactored and removed the function, as it was not completely necessary

entrypoint.sh

entrypoint.sh
ChiefHolland: Refactored

entrypoint.sh

entrypoint.sh
ChiefHolland: Refactored

CONTRIBUTING.md
```diff
@@ -0,0 +1,33 @@
...
+    pipenv run invoke goat
+    ```
+
+    Or add `--debug` for more information.
+
+2. For speed during development, `docker` caching can be helpful:
+
+    ```bash
+    docker build .
+    docker run -e RUN_LOCAL=true -it -v $PWD:/goat/ --rm <first several character of the hash output from the build step>
```
```suggestion
    docker run -e RUN_LOCAL=true -v $PWD:/goat/ --rm <first several character of the hash output from the build step>
```

README.md
```diff
@@ -103,20 +103,9 @@ with:
 
 ## Releases
 
-The `goat` project does not do releases, as it is intended as a minimum expectation that evolves over time. Please refer to `main` or, in limited
+The `goat` project does not do releases, as it is intended as a minimum expectation that evolves over time. Please refer to `main` or, in limited
```
```suggestion
The `goat` project does not do releases, as it is intended as a minimum expectation that evolves over time. Please refer to `main` or, in limited
```

JonZeolla (changes requested):

entrypoint.sh
```diff
@@ -97,40 +87,149 @@ ${overlap}"
...
+  feedback INFO "$success completed successfully"
+done
+
+if [ "${linter_failed:-true}" == "true" ]; then
+  for failure in "${linter_failures[@]}"; do
+    feedback ERROR "$failure found errors"
+  done
+  feedback ERROR "Linting failed"
+  exit 1
+else
```
```suggestion
fi
```

entrypoint.sh
```diff
@@ -97,40 +87,149 @@ ${overlap}"
...
+  feedback INFO "Linters found no errors."
```
```suggestion

feedback INFO "Linters found no errors."
```

entrypoint.sh
```diff
@@ -97,40 +87,149 @@ ${overlap}"
...
+  exit 0
```
```suggestion
```

entrypoint.sh
```diff
@@ -97,40 +87,149 @@ ${overlap}"
...
+fi
```
JonZeolla, line 231:
```suggestion
```

entrypoint.sh
```diff
@@ -97,40 +87,149 @@ ${overlap}"
...
+
```
```suggestion
```

JonZeolla (changes requested):

JonZeolla (approved):
🎉 nice work, LGTM!

PR #97, opened 2023-05-12
feat: replace super-linter with custom linter list

ChiefHolland:
# Contributor Comments

This PR removes the super-linter and replaces it with a list of linters in etc/linters.json. Configuration files for each linter were created in etc/ and match the configs contained in the super-linter. Logic was included to allow custom arguments to be passed in for each linter in an environment variable, as well as a regex-based file exclusion list.

## Pull Request Checklist


.github/workflows/commit.yml
```diff
@@ -1,7 +1,7 @@
 ---
 name: "CI"
 
-on:
+on: # yamllint disable-line rule:truthy
```
ChiefHolland, line 4: This was throwing false positives by yamllint. https://github.com/adrienverge/yamllint/issues/430

.github/workflows/security.yml
```diff
@@ -1,15 +1,15 @@
 ---
 name: "Security"
 
-on:
+on: # yamllint disable-line rule:truthy
```
ChiefHolland, line 4: Same false positive

entrypoint.sh
ChiefHolland, line 1: shfmt ate this one's lunch

.github/etc/dictionary.txt
```diff
@@ -1,2 +1,5 @@
 capricornis
 crispus
+kubeconform
+textlint
+hadolint
```
JonZeolla: Please lowercase and alphabetize; put on backlog to make this automated [here](https://www.wrike.com/open.htm?id=1118990436)

.github/workflows/commit.yml
JonZeolla: Sounds good; need to make sure this makes it to cookiecutter

Dockerfile
```diff
@@ -1,8 +1,10 @@
-# Based on python:alpine as of February 2021
-FROM github/super-linter:slim-v5.0.0
+FROM ghcr.io/yannh/kubeconform:v0.6.1 as kubeconfrm
```
JonZeolla:
```suggestion
FROM ghcr.io/yannh/kubeconform:v0.6.1 as kubeconform
```

Dockerfile
```diff
@@ -21,23 +23,41 @@ LABEL org.opencontainers.image.licenses="MIT"
 WORKDIR /etc/opt/goat/
 ENV PIP_NO_CACHE_DIR=1
 COPY Pipfile Pipfile.lock ./
+COPY --from=kubeconfrm /kubeconform /usr/bin/
```
JonZeolla:
```suggestion
COPY --from=kubeconform /kubeconform /usr/bin/
```

Dockerfile
```diff
@@ -1,8 +1,10 @@
-# Based on python:alpine as of February 2021
-FROM github/super-linter:slim-v5.0.0
+FROM ghcr.io/yannh/kubeconform:v0.6.1 as kubeconfrm
+FROM hadolint/hadolint:v2.12.1-beta-alpine as hadolint
```
JonZeolla: Do we need the beta specifically?  It seems like it just [bumped github action deps](https://github.com/hadolint/hadolint/releases) which is nice but I generally prefer to avoid betas.

Dockerfile
```diff
@@ -1,8 +1,10 @@
...
+FROM koalaman/shellcheck:v0.9.0 as shellcheck
+FROM mvdan/shfmt:v3.6.0 as shfmt
+FROM rhysd/actionlint:1.6.24 as actionlint
 
-# Required for the github/super-linter log (cannot be disabled)
-RUN mkdir -p /tmp/lint/
+FROM python:3.10-alpine3.17 as base_image
```
JonZeolla, line 6: I assume we need `python` at runtime for something?

Dockerfile
```diff
@@ -21,23 +23,41 @@ LABEL org.opencontainers.image.licenses="MIT"
...
-RUN pip install pipenv \
-    && pipenv install --deploy --ignore-pipfile \
-    && pip install shed ruff \
-    && apk upgrade \
+RUN apk upgrade \
     && apk --no-cache add go \
     jq \
     npm \
     tini \
     ruby \
```
JonZeolla: Do we need ruby and rubocop/rubocop-gitub below?  I thought those were for `super-linter`.  If so, we can remove the `gem install`s as well

Dockerfile
```diff
@@ -21,23 +23,41 @@ LABEL org.opencontainers.image.licenses="MIT"
...
     markdown-link-check \
-    && mkdir -p /opt/goat/log
+    && mkdir -p /opt/goat/log \
+    && pip install pipenv \
+    ruff \
+    mypy \
+    cfn-lint \
+    black \
+    yamllint \
+    && pipenv install --deploy --ignore-pipfile
```
JonZeolla: Why are we doing a pip install before if we're doing a pipenv install here?  We could just add the above to the `Pipfile` of the parent project in the `[packages]` section, then we will get version pinned and locked installs with auto updating via the `goat`'s auto updater

JonZeolla: We will need to do `pipenv run` them at runtime, which could be done with the existing `executor` feature in `linters.json`

entrypoint.sh
```diff
@@ -7,227 +7,247 @@ set -o pipefail
 
 # shellcheck disable=SC2034
 {
-  declare -r ERROR='\033[0;31m'
```
JonZeolla, line 10: I don't know if I can handle all these tabs 👀

entrypoint.sh
```diff
@@ -7,227 +7,247 @@ set -o pipefail
...
+
+	# Set workspace to /goat/ for local runs
+	export DEFAULT_WORKSPACE="/goat"
+
+	# Create variables for the various dictionary file paths
+	export GLOBAL_DICTIONARY="/etc/opt/goat/seiso_global_dictionary.txt"
+	export REPO_DICTIONARY="${GITHUB_WORKSPACE:-/goat}/.github/etc/dictionary.txt"
+
+	# Map certain environment variables
+	if [[ ${INPUT_DISABLE_TERRASCAN-} == "true" ]]; then
```
JonZeolla: We aren't including terrascan, right?  We can just remove this

entrypoint.sh
```diff
@@ -7,227 +7,247 @@ set -o pipefail
...
+	if [[ -n ${INPUT_EXCLUDE:+x} ]]; then
+		export FILTER_REGEX_EXCLUDE="${INPUT_EXCLUDE}"
+	fi
+
+	if [[ ${INPUT_LOG_LEVEL:='VERBOSE'} =~ ^(ERROR|WARN|NOTICE|VERBOSE|DEBUG|TRACE)$ ]]; then
+		export LOG_LEVEL="${INPUT_LOG_LEVEL}"
+		export ACTIONS_RUNNER_DEBUG="true"
+	fi
+
+	linter_failed="false"
```
JonZeolla: I think this is a mistake from before, we should default closed

entrypoint.sh
```diff
@@ -7,227 +7,247 @@ set -o pipefail
...
+	return 1
+}
 
-  fi
+function detect_cloudformation_file() {
+	local file="$1"
 
-  # When run in a pipeline, move per-repo configurations into the right location at runtime so super-linter finds them, overwriting the defaults.
-  # This will handle hidden and non-hidden files, as well as sym links
-  cp -p "${GITHUB_WORKSPACE:-.}/.github/linters/"* "${GITHUB_WORKSPACE:-.}/.github/linters/".* /etc/opt/goat/ || true
```
JonZeolla, line 107: Are we removing support for per-repo linter configs?  I believe we depend on this today.

entrypoint.sh
```diff
@@ -7,227 +7,247 @@ set -o pipefail
...
+		grep -q -E '(kind):' "${file}"; then
+		return 0
+	fi
 
-  # Turn off the possum
-  export SUPPRESS_POSSUM="true"
-
-  if [[ -n ${GITHUB_WORKSPACE:-} ]]; then
-    echo "Setting ${GITHUB_WORKSPACE} as safe directory"
-    git config --global --add safe.directory "${GITHUB_WORKSPACE}"
```
JonZeolla, line 101: This was previously required regardless of `super-linter`.

JonZeolla: https://github.com/SeisoLLC/goat/pull/62

entrypoint.sh
```diff
@@ -7,227 +7,247 @@ set -o pipefail
...
 
-function super_lint() {
-  superlinter_logfile="/opt/goat/log/super-linter.log"
+function detect_kubernetes_file() {
+	local file="$1"
 
-  echo -e "\nRunning Super-Linter\n--------------------------\n"
-  echo "===============================" >> "$superlinter_logfile"
-  echo "SUPER-LINTER" >> "$superlinter_logfile"
+	if grep -q -v 'kustomize.config.k8s.io' "${file}" &&
```
JonZeolla, line 88: Can you please add a comment to explain why we want this?

entrypoint.sh
```diff
@@ -7,227 +7,247 @@ set -o pipefail
...
+	local file="$1"
 
-  # When run in a pipeline, move per-repo configurations into the right location at runtime so super-linter finds them, overwriting the defaults.
-  # This will handle hidden and non-hidden files, as well as sym links
-  cp -p "${GITHUB_WORKSPACE:-.}/.github/linters/"* "${GITHUB_WORKSPACE:-.}/.github/linters/".* /etc/opt/goat/ || true
-  
-  set +e
-  superlinter_result=$(/action/lib/linter.sh >> "$superlinter_logfile" 2>&1; echo $?)
-  set -e
+	if grep -q 'AWSTemplateFormatVersion' "${file}" >/dev/null; then
```
JonZeolla, line 105: Please add a comment to explain why this is here

entrypoint.sh
```diff
@@ -7,227 +7,247 @@ set -o pipefail
...
+		return 0
+	fi
 
-  echo "-------------------------------" >> "$superlinter_logfile"
+	if grep -q -E '(AWS|Alexa|Custom)::' "${file}" >/dev/null; then
```
JonZeolla, line 110: Please add a comment to explain why this is here

entrypoint.sh
```diff
@@ -7,227 +7,247 @@ set -o pipefail
...
-      cmd="${linter[executor]} ${linter[name]} ${linter[args]} ${file}"
-    else
-      cmd="${linter[name]} ${linter[args]} ${file}"
-    fi
-    eval "$cmd" >> "${linter[logfile]}" 2>&1
-  done
+	local linter_args="${linter[args]}"
+
+	if [[ -v "${linter[env]}" && -n "${!linter[env]}" ]]; then
+		linter_args="${!linter[env]}"
```
JonZeolla, line 154: Are you purposefully overwriting here instead of being additive?

entrypoint.sh
```diff
@@ -7,227 +7,247 @@ set -o pipefail
...
+	fi
+
+	for type in "${linter_filetypes[@]}"; do
+		if [[ $type == "all" ]]; then
+			cmd="${linter[name]} $linter_args"
+			eval "$cmd" >>"${linter[logfile]}" 2>&1
```
JonZeolla, line 160:
```suggestion
			eval "$cmd" >> "${linter[logfile]}" 2>&1
```

etc/actionlint.yml
```diff
@@ -0,0 +1,5 @@
+---
+self-hosted-runner:
```
JonZeolla: We use gh hosted runners.  According to [this](https://github.com/rhysd/actionlint/blob/94c4f0119fdc7f776c7a472a981121cafd17a3cd/docs/config.md#configuration-file) shouldn't this just be empty then?

etc/linters.json
```diff
@@ -2,27 +2,152 @@
...
+        ],
+        "installer": "npm",
+        "executor": "npx",
+        "env": "JSCPD_CONFIG"
+    },
+    {
+        "name": "actionlint",
+        "args": "-config-file /etc/opt/goat/actionlint.yml",
+        "filetype": [
+            "all"
```
JonZeolla: Aren't these just `.yml` or `.yaml` file? From [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#about-yaml-syntax-for-workflows):
> Workflow files use YAML syntax, and must have either a `.yml` or `.yaml` file extension. If you're new to YAML and want to learn more, see "[Learn YAML in Y minutes](https://learnxinyminutes.com/docs/yaml/)."

entrypoint.sh
```diff
@@ -7,227 +7,247 @@ set -o pipefail
...
+	done
+
+	for file in $(get_files_matching_filetype "${included[@]}"); do
+		if [[ "${linter[executor]+x}" ]]; then
+			cmd="${linter[executor]} ${linter[name]} $linter_args ${file}"
+		else
+			cmd="${linter[name]} $linter_args ${file}"
+		fi
+
+		eval "$cmd" >>"${linter[logfile]}" 2>&1
```
JonZeolla, line 179:
```suggestion
		eval "$cmd" >> "${linter[logfile]}" 2>&1
```

entrypoint.sh
JonZeolla: Also, need to update the `README.md` to be accurate since it describes the terrascan input config

entrypoint.sh
```diff
@@ -7,227 +7,247 @@ set -o pipefail
...
+	# Create variables for the various dictionary file paths
+	export GLOBAL_DICTIONARY="/etc/opt/goat/seiso_global_dictionary.txt"
+	export REPO_DICTIONARY="${GITHUB_WORKSPACE:-/goat}/.github/etc/dictionary.txt"
+
+	# Map certain environment variables
+	if [[ ${INPUT_DISABLE_TERRASCAN-} == "true" ]]; then
+		export VALIDATE_TERRAFORM_TERRASCAN="false"
+	fi
+
+	if [[ ${INPUT_DISABLE_MYPY-} == "true" ]]; then
```
JonZeolla, line 45: Does this actually disable mypy?  I don't think it does, does it?  It seems like we're missing a disable linter feature of `linters.json` and `entrypoint.sh`; if the provided env var depending on the linter, something like `INPUT_DISABLE_${linter[name]}`, has a value of `true` then log (or queue) a message that it was disabled and return with exit 0.

CONTRIBUTING.md
```diff
@@ -28,6 +28,59 @@ There are two ways of running the `goat` locally:
     docker run -e RUN_LOCAL=true -v $PWD:/goat/ --rm <first several character of the hash output from the build step>
```
JonZeolla: `RUN_LOCAL` is no longer needed, right?
```suggestion
    docker run -v "$PWD:/goat/" --rm <first several character of the hash output from the build step>
```

CONTRIBUTING.md
```diff
@@ -28,6 +28,59 @@ There are two ways of running the `goat` locally:
...
     ```
 
+3. To pass in custom configs for individual linters or to exclude files with regex, set environment variables:
+
+    ```bash
+    docker run -e RUN_LOCAL=true -e INPUT_EXCLUDE='.*\.json$' -e BLACK_CONFIG='--required-version 21.9b0' -v $PWD:/goat/ --rm <hash>
```
```suggestion
    docker run -e INPUT_EXCLUDE='.*\.json$' -e BLACK_CONFIG='--required-version 21.9b0' -v "$PWD:/goat/" --rm <hash>
```

JonZeolla (changes requested):

Dockerfile
JonZeolla: It looks like yes because we will do `pipenv install` and `pipenv run`

entrypoint.sh
JonZeolla: Either way, that should be documented in the `README.md`

entrypoint.sh
JonZeolla: I confirmed; it doesn't

entrypoint.sh
```diff
@@ -7,227 +7,247 @@ set -o pipefail
...
-  export SUPPRESS_POSSUM="true"
-
-  if [[ -n ${GITHUB_WORKSPACE:-} ]]; then
-    echo "Setting ${GITHUB_WORKSPACE} as safe directory"
-    git config --global --add safe.directory "${GITHUB_WORKSPACE}"
+	return 1
+}
 
-  fi
+function detect_cloudformation_file() {
```
JonZeolla, line 98: Document what return 0 vs 1 means

entrypoint.sh
```diff
@@ -7,227 +7,247 @@ set -o pipefail
...
 ${overlap}"
-    feedback ERROR "Overlap was detected in the per-repo and global dictionaries"
-  fi
+		feedback ERROR "Overlap was detected in the per-repo and global dictionaries"
+	fi
 }
 
-function super_lint() {
-  superlinter_logfile="/opt/goat/log/super-linter.log"
+function detect_kubernetes_file() {
```
JonZeolla, line 82: Document what return 0 vs 1 means.  I believe 1 represents a failure, not a match, right?  1 is a truthy value and many would read `return 1` as `return True` like, something that matches was detected.  But I don't believe that's how this bash works; I'm pretty sure it sees it as a non-zero return so it failed

entrypoint.sh
```diff
@@ -7,227 +7,247 @@ set -o pipefail
...
+	matching_files=()
+
+	for file in "${filenames[@]}"; do
+		filename=$(basename "$file")
+
+		for filetype in "${linter_filetypes[@]}"; do
+			if [[ $filename == *"$filetype" ]]; then
+				if [ "${linter[name]}" == "cfn-lint" ]; then
+					if ! detect_cloudformation_file "${file}"; then
+						break
```
JonZeolla: Are we sure we want to `break` here and not `continue`?  Same with below 2.  If there are yml and yaml files that match, wouldn't it miss the second filetype extension?

Dockerfile
ChiefHolland: Changed to v2.12.0-alpine

.github/etc/dictionary.txt
ChiefHolland: Done

Dockerfile
ChiefHolland: Done

JonZeolla (changes requested):

entrypoint.sh
ChiefHolland: Re-added

Dockerfile
ChiefHolland: Done

entrypoint.sh
ChiefHolland: Re-added

entrypoint.sh
ChiefHolland: Done and added conditional logic for the mypy linter.

entrypoint.sh
ChiefHolland: This was removed in favor of checking to see if the linter_failures array contained any values.

entrypoint.sh
ChiefHolland: This was how super-linter designated it, but I agree, it doesn't seem to match typical boolean use. I flipped the values in both functions.

entrypoint.sh
ChiefHolland: This was how super-linter designated it, but I agree, it doesn't seem to match typical boolean use. I flipped the values in both functions.

entrypoint.sh
ChiefHolland: True. Changed to continue

CONTRIBUTING.md
ChiefHolland: Removed

CONTRIBUTING.md
ChiefHolland: Removed

entrypoint.sh
ChiefHolland: Done

entrypoint.sh
ChiefHolland: Yes, we want to overwrite the default arguments with the supplied ones. 
Documented

entrypoint.sh
ChiefHolland: Done

entrypoint.sh
ChiefHolland: Done

etc/actionlint.yml
ChiefHolland: Done

etc/linters.json
ChiefHolland: Done

entrypoint.sh
ChiefHolland: Added logic for disabling mypy

entrypoint.sh
JonZeolla: "Less confusing than the super-linter" should go in our README.md 🤣

I'm actually serious though, but no need to call them out directly.  Maybe something like "Designed to be easier to understand, and be more opinionated than other open source projects"

entrypoint.sh
JonZeolla: Yeah, true.  It also mirrors what we do in easy_infra, so I'll be replacing the approach there as well with something better (albeit more complicated)

entrypoint.sh
ChiefHolland: Added it to the README.

Dockerfile
```diff
@@ -21,21 +23,32 @@ LABEL org.opencontainers.image.licenses="MIT"
...
 RUN pip install pipenv \
-    && pipenv install --deploy --ignore-pipfile \
-    && pip install shed ruff \
+    && pipenv install --system --deploy --ignore-pipfile \
     && apk upgrade \
     && apk --no-cache add go \
     jq \
     npm \
     tini \
     ruby \
```
JonZeolla: You missed the base `ruby`
```suggestion
```

Pipfile
```diff
@@ -4,6 +4,11 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
+ruff = "==0.0.267"
```
JonZeolla: Don't version pin them here, that will happen automatically in the lock file.  If you pin it here then the auto-updates will never find anything newer.  Just use this across the board
```suggestion
ruff = "*"
```

README.md
```diff
@@ -1,6 +1,7 @@
 # Grand Opinionated AutoTester (GOAT)
 
 The Grand Opinionated AutoTester (GOAT) automatically applies Seiso's standard testing.
+This project has been designed to be easier to understand and more opinionated than other open source projects.
```
```suggestion
This project has been designed to be easier to understand and more opinionated than other open source multi-linter project.
```

entrypoint.sh
```diff
@@ -7,227 +7,277 @@ set -o pipefail
...
+						continue
+					fi
+				fi
+				if [ "${linter[name]}" == "kubeconform" ]; then
+					if detect_kubernetes_file "${file}"; then
+						continue
+					fi
+				fi
+				if [ "${linter[name]}" == "actionlint" ]; then
+					local action_path="./.github/workflows/"
```
JonZeolla: Shouldn't this be:
```suggestion
					local action_path="${GITHUB_WORKSPACE:-.}/.github/workflows/"
```

JonZeolla: It will default to ./ but also use `${GITHUB_WORKSPACE}` if it's set, such as in a pipeline, and has the added benefit that it's an absolute path when in a pipeline and doesn't care about other `cd`s or `pushd`s

etc/linters.json
```diff
@@ -2,27 +2,158 @@
...
+        "env": "SHELLCHECK_CONFIG"
+    },
+    {
+        "name": "yamllint",
+        "args": "-c /etc/opt/goat/.yamllint.yml",
+        "filetype": [
+            ".yaml",
+            ".yml"
+        ],
+        "installer": "pip",
```
JonZeolla: Isn't the installer technically `pipenv`, i.e. `pipenv install`?  Does it really matter, though?  Same for the others here; I don't know where the installer matters.

entrypoint.sh
```diff
@@ -7,227 +7,277 @@ set -o pipefail
...
+
+	for file in "${files_to_lint[@]}"; do
+		if [[ "${linter[executor]+x}" ]]; then
+			cmd="${linter[executor]} ${linter[name]} $linter_args ${file}"
+		else
+			cmd="${linter[name]} $linter_args ${file}"
+		fi
+
+		echo "$cmd" >>"${linter[logfile]}"
+		eval "$cmd" >>"${linter[logfile]}" 2>&1
```
JonZeolla, line 179: Instead of this eval and echo, can't se just use `tee` to both send it to stdout and to a file?  You just need to do it something like this (untested):
```suggestion
		eval "$cmd" 2>&1 | tee -a "${linter[logfile]}"
```

The `tee -a` appends to an existing file and may not be needed.  If so, you can just remove the `-a`

JonZeolla (changes requested):
Quick review; will need one more full run through of the actual business logic but wanted to get you this feedback sooner

Dockerfile
ChiefHolland: Done

Pipfile
ChiefHolland: Done

README.md
ChiefHolland: Done

entrypoint.sh

etc/linters.json
ChiefHolland: I'm going to remove the installer member. I was only using that to handle dynamic installation in the Dockerfile. Since we're not doing that, I don't see the need, either.

entrypoint.sh
ChiefHolland: The goal was to suppress all output during parallel processes to keep from cluttering the output. The command you suggested spits out info to stdout without context. I'd like to keep this one in for now, and revisit a more elegant solution, if that's okay. The syntax of $cmd is only being sent to the log file for debugging purposes. Now that the issue with kubeconform has been resolved, we can remove the echo line, although it might still have value in debugging.

entrypoint.sh
ChiefHolland: Done

README.md
ChiefHolland: Done

entrypoint.sh
```diff
@@ -7,227 +7,277 @@ set -o pipefail
...
-  echo -e "\nScanned ${#included[@]} files"
-  echo -e "Excluded ${#excluded[@]} files\n"
+	echo -e "\nRunning Seiso Linter\n--------------------------\n"
+
+	if [[ -n ${GITHUB_WORKSPACE:-} ]]; then
+		echo "Setting ${GITHUB_WORKSPACE} as safe directory"
+		git config --global --add safe.directory "${GITHUB_WORKSPACE}"
+	fi
+
+	# When run in a pipeline, move per-repo configurations into the right location at runtime so super-linter finds them, overwriting the defaults.
```
```suggestion
	# When run in a pipeline, move per-repo configurations into the right location at runtime so the goat finds them, overwriting the defaults.
```

entrypoint.sh
```diff
@@ -7,227 +7,277 @@ set -o pipefail
...
+				done <<<"$value"
+				continue
+			fi
+			value=$(echo "$value" | tr -d "'" | tr -d '"')
+			linter["$key"]=$value
+		done < <(echo "$line" | jq -r 'to_entries|map("\(.key)=\(.value|tojson)")|.[]')
+
+		linter[logfile]="/opt/goat/log/${linter[name]}.log"
+
+		if [[ -v VALIDATE_PYTHON_MYPY && "${VALIDATE_PYTHON_MYPY,,}" == "false" ]] && [[ "${linter[name]}" == "mypy" ]]; then
```
```suggestion
		if [[ -v VALIDATE_PYTHON_MYPY && "${VALIDATE_PYTHON_MYPY,,}" == "false" && "${linter[name]}" == "mypy" ]]; then
```

entrypoint.sh
```diff
@@ -7,227 +7,277 @@ set -o pipefail
```
JonZeolla, line 7: I have no explanation for why this happened.  Note the # of scanned files between the two.  I blame the `.git` folder for black magic:

```bash
$ docker run -it -v $(pwd):/goat seiso/goat:c230b42

Running Seiso Linter
--------------------------

Running linter: ruff
Running linter: hadolint
Running linter: kubeconform
Running linter: dockerfile_lint
Running linter: markdown-link-check
Running linter: cspell
Running linter: cfn-lint
Running linter: jscpd
Running linter: actionlint
Running linter: markdownlint
Running linter: textlint
Running linter: black
Running linter: mypy
Running linter: shfmt
Running linter: shellcheck
Running linter: yamllint

Scanned 277 files
Excluded 0 files

INFO:  shellcheck completed successfully
INFO:  hadolint completed successfully
INFO:  ruff completed successfully
INFO:  kubeconform completed successfully
INFO:  jscpd completed successfully
INFO:  yamllint completed successfully
INFO:  shfmt completed successfully
INFO:  actionlint completed successfully
INFO:  black completed successfully
INFO:  cspell completed successfully
INFO:  dockerfile_lint completed successfully
INFO:  markdownlint completed successfully
INFO:  mypy completed successfully
INFO:  markdown-link-check completed successfully
INFO:  cfn-lint completed successfully
INFO:  textlint completed successfully
INFO:  Linters found no errors.
```

```bash

$ docker run -it -v $(pwd):/goat seiso/goat:c230b42

Running Seiso Linter
--------------------------

Running linter: ruff
Running linter: hadolint
Running linter: kubeconform
Running linter: dockerfile_lint
Running linter: markdown-link-check
Running linter: cspell
Running linter: cfn-lint
Running linter: jscpd
Running linter: actionlint
Running linter: markdownlint
Running linter: textlint
Running linter: black
Running linter: mypy
Running linter: shfmt
Running linter: shellcheck
Running linter: yamllint

Scanned 482 files
Excluded 0 files

INFO:  jscpd completed successfully
INFO:  hadolint completed successfully
INFO:  ruff completed successfully
INFO:  markdown-link-check completed successfully
INFO:  yamllint completed successfully
INFO:  actionlint completed successfully
INFO:  textlint completed successfully
INFO:  cspell completed successfully
INFO:  black completed successfully
INFO:  mypy completed successfully
INFO:  shellcheck completed successfully
INFO:  kubeconform completed successfully
INFO:  shfmt completed successfully
INFO:  markdownlint completed successfully
INFO:  dockerfile_lint completed successfully
INFO:  cfn-lint completed successfully
INFO:  Linters found no errors.
```

```bash
$ find . | wc -l
    1267
```

JonZeolla (changes requested):
Sorry, I need to un-approve just because it introduces breaking changes when I tested it against some projects.

It scans easy_infra in 7 seconds 👍👍, but fails the build due to the shfmt tabs and cfn-lint issues; we need a way to turn these off or loosen the shfmt rules for now.

We may be able to add an `.editorconfig` file to the `goat/etc/` folder and configure it in line with https://github.com/mvdan/sh/blob/master/cmd/shfmt/shfmt.1.scd but I can't get it to work.

I suggest scanning some other projects manually and see what happens

Also, when we skip a linter we need to add a WARNING: to the end that it was skipped (I did this in 5f92079c42c2893959b66e8a999630732eed1779)

JonZeolla:
I don't know how to get it to find the .editorconfig because they don't seem to have a way to pass an arbitrary location to load it from. It loads from current directory up to the root - see https://editorconfig.org/#file-location

That will also cause an issue for the `.github/linters/` config override approach.

Perhaps the easiest approach is just remove `shfmt` for now and put it on the backlog?

BTW it looks like there was a config for this in `super-linter` [here](https://github.com/super-linter/super-linter/blob/0e9d0d366114dc7f3476500b8a74884e0ba78b30/.editorconfig#L10-L12).

JonZeolla:
We don't need to change `cfn-lint`; I made the recommended changes

JonZeolla (approved):
LGTM! 🚀

PR #105, opened 2023-05-26
feat: refactor reformat task

ChiefHolland:
# Contributor Comments

This PR includes two minor fixes for the textlint linter and a refactoring of the reformat function in tasks.py. 

The textlint fixes include the following:

1. Remove .rst support. There is an installation problem with a [dependent package](https://github.com/jimo1001/docutils-ast-writer) that prevents the rst plugin from operating.
2. Remove globbing from the command arguments in favor of a per-filetype option. This was necessary in order to properly implement the auto-fix command line option.

Formerly the reformat function in tasks.py called two specific functions (black and isort) from the goat docker container and reformatted files in the repo. The changes introduced here are in alignment with centralizing the list of linters and their options in linters.json. A new field was added to those linter objects that have a builtin auto-fix command line argument, which is populated with the necessary arguments to enable that feature. There is now an environment variable that triggers which set of arguments to use. The python task now passes the environment variable to the entrypoint.sh script to enable reformatting.

## Pull Request Checklist


CONTRIBUTING.md
```diff
@@ -80,6 +80,11 @@ There are two ways of running the `goat` locally:
     Note: Linter env variables must be formatted as <LINTER_CONFIG>, i.e. RUFF_CONFIG, CFN_LINT_CONFIG,
     etc., and the values supplied will overwrite the default arguments supplied in the goat.
 
+4. Autofix is available for certain linters and can be invoked in two ways:
```
JonZeolla, line 83: Remember, this is a GitHub action. Document the GitHub action method to do this. You will need to map an INPUT_AUTO_FIX to AUTO_FIX like we’ve done otherwise

CONTRIBUTING.md
```diff
@@ -80,6 +80,11 @@ There are two ways of running the `goat` locally:
...
+   1. `docker run -e AUTO_FIX="true" -v $PWD:/goat/ -rm <hash>`
+   2. `pipenv run invoke reformat`
+
 ### Note
```
JonZeolla: This is becoming a linter updates section or similar. Make heading more precise based on contents please

entrypoint.sh
```diff
@@ -228,6 +228,16 @@ function seiso_lint() {
 			linter["$key"]=$value
 		done < <(echo "$line" | jq -r 'to_entries|map("\(.key)=\(.value|tojson)")|.[]')
 
+		if [[ ${AUTO_FIX-} == "true" ]]; then
```
JonZeolla: Why not :- see https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

etc/linters.json
```diff
@@ -89,12 +90,13 @@
     },
     {
         "name": "textlint",
-        "args": "-c /etc/opt/goat/.textlintrc \"/goat/**\"",
+        "args": "-c /etc/opt/goat/.textlintrc --dry-run",
```
JonZeolla: Why dry run?

tasks.py
```diff
@@ -258,26 +258,25 @@ def reformat(_c, debug=False):
     if debug:
         getLogger().setLevel("DEBUG")
 
-    entrypoint_and_command = [
-        ("isort", ". --settings-file /action/lib/.automation/.isort.cfg"),
-        ("black", "."),
-    ]
     image = "seiso/goat:latest"
+    environment = {}
+    environment["AUTO_FIX"] = "true"
```
JonZeolla: Input

entrypoint.sh
```diff
@@ -258,6 +268,9 @@ function seiso_lint() {
 			cat "/opt/goat/log/${pids[$p]}.log"
 			linter_failures+=("${pids[$p]}")
 		else
+			if [[ ${AUTO_FIX-} == "true" ]]; then
```
JonZeolla: :- again

JonZeolla (changes requested):
Quick feedback from my mobile to keep this moving. Thanks!

JonZeolla:
- [x] make sure to update the readme(s) on how to use this. Think about GitHub action use cases (unlikely valuable), and taskfile usage.
- [ ] update the taskfiles themselves to use this approach in a reformat cmd

CONTRIBUTING.md
ChiefHolland: Created the map

CONTRIBUTING.md
ChiefHolland: Done

entrypoint.sh

etc/linters.json
ChiefHolland: Dug back into the linter docs. Yep, unnecessary. Removed.

tasks.py
ChiefHolland: Done

entrypoint.sh

tasks.py
```diff
@@ -299,7 +297,7 @@ def update(_c, debug=False):
     if debug:
         getLogger().setLevel("DEBUG")
 
-    repo = "github/super-linter"
+    repo = ""  # TODO Repurpose update to manage new dependencies
```
ChiefHolland, line 300: I wasn't sure how you wanted to handle this one. The update task and functions below are only used to update super-linter. I created a task to refactor these functions to handle updating the new dependencies in the multi-stage approach, which will go into the next sprint. I also planned on tackling the move to Taskfile in that sprint, so refactoring this might be moot.

JonZeolla: Yeah, don't refactor anything here or make substantial changes.  Just map it to the taskfile work.  We want to migrate from tasks.py to Taskfile.yml with no changes, then after that we make the changes in the Taskfile.  Two step process to avoid a refactor + logic change which is hard to review/get right

JonZeolla: This should be changed to auto-update the `python:3.10-alpine3.17` base that we're using.

README.md
```diff
@@ -90,6 +90,14 @@ jobs:
       line_length: 120
     EOF
     ```
+  
+1. Autofix code formatting errors using those linters with a built-in `fix` option.
```
```suggestion
1. Autofix code formatting errors using those linters with a built-in `fix` option. Note, this will create a `dirty` git directory, and you will need to manage committing and pushing any changes; consider setting `INPUT_AUTO_FIX=true` locally instead.
```

entrypoint.sh
```diff
@@ -228,6 +232,16 @@ function seiso_lint() {
 			linter["$key"]=$value
 		done < <(echo "$line" | jq -r 'to_entries|map("\(.key)=\(.value|tojson)")|.[]')
 
+		if [[ ${AUTO_FIX:-} == "true" ]]; then
+			if [[ -v linter[autofix] && -n "${linter[autofix]}" ]]; then
+				linter[args]="${linter[autofix]}"
```
JonZeolla, line 242:
```suggestion
				# Replacing the linter's args with the autofix args for that linter
				linter[args]="${linter[autofix]}"
```

JonZeolla: Took me a minute of looking at this to figure out what was happening so suggest a comment

entrypoint.sh
```diff
@@ -228,6 +232,16 @@ function seiso_lint() {
...
+			else
+				echo "${linter[name]} has no autofix option and has been skipped"
```
JonZeolla, line 244: I don't think this is true; it's still running the linter, just with the default args (i.e. not fixing).  I think we need to unset some things as well for this to be true.  I tested it by building `seiso/goat:ca7b334` and injecting a Hadolint failure;  it failed the Hadolint linter, with no skip message shown.

JonZeolla (changes requested):

entrypoint.sh
ChiefHolland: @JonZeolla, The linter array is unset on each pass in the while loop. Do you think the bug relate to the fact I'm declaring it as a global variable? Because I'm running a background process on a declared iteration of the linter array, is it registering that linter[autofix] is a valid key even though the current pass doesn't have that value? If that's plausible, I can make it a local variable and pass it from function to function. 

And the python image is not released in the same way super-linter is. To update that base image would take a refactor of the update functions.

ChiefHolland: I refactored the global variable AA to a local one. Adding an error test file now skips the linter when an autofix arg is not present.

tasks.py
ChiefHolland: I updated to the latest 3.10-alpine3.18 version, but docker-library, the supporters of this image, do not have releases. Figuring out how to check for new base images and auto-updating would require refactoring the update functions. Also we're locked in at Python 3.10 for now. 3.11 is a different code path, so if we migrate up to a new Python version, we'd have to update this code, again. I'll take your lead on this one. Point me in a direction, and I'll shoot, but I move whatever we do, we do it in a subsequent sprint and push this one through.

entrypoint.sh
```diff
@@ -240,7 +258,7 @@ function seiso_lint() {
 		echo "Running linter: ${linter[name]}"
 		echo "${linter[name]^^}" >>"${linter[logfile]}"
 
-		lint_files &
+		lint_files linter "${linter_filetypes[@]}" &
```
JonZeolla, line 263:
```suggestion
		# The string "linter" gets dereferenced back into a variable on the receiving end
		lint_files linter "${linter_filetypes[@]}" &
```

entrypoint.sh
```diff
@@ -115,69 +119,72 @@ function detect_cloudformation_file() {
...
+			matching_files+=("${file}")
+		fi
 	done
 
 	echo "${matching_files[@]}"
 }
 
 function lint_files() {
-	local linter_args="${linter[args]}"
+	local -n linter_array="$1"
```
JonZeolla, line 156:
```suggestion
	# Turn the received string back into an object
	local -n linter_array="$1"
```

JonZeolla (dismissed):
LGTM!

Task/python/Taskfile.yml
```diff
@@ -59,11 +59,11 @@ tasks:
       - sh: docker pull seiso/goat:latest || echo "Failed to pull the latest goat, continuing anyway..."
     cmds:
       - >
-        docker run --rm -v "${PWD}:/goat" -e RUN_LOCAL=true
+        docker run --rm -v "${PWD}:/goat"
```
JonZeolla:
```suggestion
        docker run --rm -v "${PWD}:/goat" -e INPUT_AUTO_FIX=true seiso/goat:latest
```

Task/python/Taskfile.yml
```diff
@@ -59,11 +59,11 @@ tasks:
...
         --entrypoint isort seiso/goat:latest .
```
JonZeolla:
```suggestion
```

Task/python/Taskfile.yml
```diff
@@ -59,11 +59,11 @@ tasks:
...
         --settings-file /action/lib/.automation/.isort.cfg
```
JonZeolla:
```suggestion
```

Task/python/Taskfile.yml
```diff
@@ -59,11 +59,11 @@ tasks:
...
       - >
```
JonZeolla:
```suggestion
```

Task/python/Taskfile.yml
```diff
@@ -59,11 +59,11 @@ tasks:
...
-        docker run --rm -v "${PWD}:/goat" -e RUN_LOCAL=true
+        docker run --rm -v "${PWD}:/goat"
```
JonZeolla:
```suggestion
```

Task/python/Taskfile.yml
```diff
@@ -59,11 +59,11 @@ tasks:
...
         --entrypoint black seiso/goat:latest .
```
JonZeolla:
```suggestion
```

JonZeolla (commented):
Tweaks to taskfiles

Task/python/Taskfile.yml
```diff
@@ -59,12 +59,7 @@ tasks:
       - sh: docker pull seiso/goat:latest || echo "Failed to pull the latest goat, continuing anyway..."
     cmds:
       - >
```
JonZeolla:
```suggestion
      - docker run --rm -v "${PWD}:/goat" -e INPUT_AUTO_FIX=true seiso/goat:latest
```

Task/python/Taskfile.yml
```diff
@@ -59,12 +59,7 @@ tasks:
...
-        docker run --rm -v "${PWD}:/goat" -e RUN_LOCAL=true
-        --entrypoint isort seiso/goat:latest .
-        --settings-file /action/lib/.automation/.isort.cfg
-      - >
-        docker run --rm -v "${PWD}:/goat" -e RUN_LOCAL=true
-        --entrypoint black seiso/goat:latest .
+        docker run --rm -v "${PWD}:/goat" -e INPUT_AUTO_FIX=true seiso/goat:latest
```
JonZeolla:
```suggestion
```

JonZeolla (commented):
More taskfile tweaks

JonZeolla (commented):
More taskfile tweaks

JonZeolla (approved):
🚀

PR #107, opened 2023-06-07
fix: file exclusion regex not applied for linters marked "all"

ChiefHolland:
# Contributor Comments

This PR fixes a bug in which linters in the linters.json file that were marked to check all filetypes were scanning all files in the directory instead of just those in the `included` array. 

## Pull Request Checklist

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -70,6 +70,7 @@ function check_environment() {
 		mainline="${GITHUB_BASE_REF##*/}"
 		if [[ ${mainline} != "main" ]]; then
 			feedback ERROR "Base branch name is not main"
+			exit 1
 		fi
 	fi
 
@@ -80,6 +81,13 @@ function check_environment() {
 		feedback WARNING "The following words are already in the global dictionary:
 ${overlap}"
 		feedback ERROR "Overlap was detected in the per-repo and global dictionaries"
+		exit 1
+	fi
+
+	# Ensure dictionaries are sorted
+	if ! sort -c "${REPO_DICTIONARY}" 2>/dev/null; then
+		feedback ERROR "The repo dictionary must be sorted"
+		exit 1
 	fi
 }
 
@@ -171,7 +179,7 @@ function lint_files() {
 		fi
 
 		files_to_lint="$(get_files_matching_filetype "$type" "${linter_array[name]}" "${included[@]}")"
-		
+
 		if [ "${#files_to_lint}" -eq 0 ]; then
 			return
 		fi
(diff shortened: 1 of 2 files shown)


entrypoint.sh
```diff
@@ -165,13 +173,13 @@ function lint_files() {
 
 	for type in "${filetypes_to_lint[@]}"; do
 		if [[ $type == "all" ]]; then
-			cmd="${linter_array[name]} $linter_args"
+			cmd="${linter_array[name]} $linter_args $(printf '%q ' "${included[@]}")"
```
ChiefHolland, line 176: I've rarely, if ever, used printf in bash. Adding the array as ${included[*]} would work but was buggy in some local testing, as it would error when the array contain file names as `./file` . The `.` would break it. ${included[@]} worked, except hadolint had a problem assigning an array to the string var `cmd`. ChatGPT gave me the syntax above, and it worked in all of my local testing on the monopreso and goat folders.

entrypoint.sh
```diff
@@ -207,11 +215,11 @@ function seiso_lint() {
 	included=()
 
 	while read -r file; do
-		if [[ -n ${INPUT_EXCLUDE:+x} && "${file}" =~ ${INPUT_EXCLUDE} ]]; then
+		if [[ -n ${FILTER_REGEX_EXCLUDE:+x} && "${file}" =~ ${FILTER_REGEX_EXCLUDE} ]]; then
```
ChiefHolland, line 218: We were mapping these variables earlier in the script, so I changed them here for consistency with other vars.

entrypoint.sh
```diff
@@ -207,11 +215,11 @@ function seiso_lint() {
...
 			excluded+=("${file}")
 			continue
 		fi
-		included+=("$file")
+		included+=("${file}")
```
ChiefHolland, line 222: Also a consistency change with how I was adding to `excluded=()`

etc/linters.json
```diff
@@ -64,7 +64,7 @@
     },
     {
         "name": "jscpd",
-        "args": "--config /etc/opt/goat/.jscpd.json *",
```
ChiefHolland, line 67: Yep, that surely won't handle any exclusions.

entrypoint.sh
JonZeolla: This is a good use case for printf. Hats off to the llm

etc/linters.json
JonZeolla: Ah good catch

JonZeolla (approved):
🤘

PR #108, opened 2023-06-14
feat: automate autofix

ChiefHolland:
# Contributor Comments

These changes enable the goat's auto-fix feature by default when running locally and disable it when running in a pipeline. There is an override variable that can force either condition in either environment (`INPUT_AUTO_FIX`).

## Pull Request Checklist

### DIFF
--- a/.github/etc/dictionary.txt
+++ b/.github/etc/dictionary.txt
@@ -3,5 +3,4 @@ capricornis
 crispus
 hadolint
 kubeconform
-noreformat
 textlint
--- a/.github/etc/dictionary.txt
+++ b/.github/etc/dictionary.txt
@@ -3,4 +3,5 @@ capricornis
 crispus
 hadolint
 kubeconform
+noreformat
 textlint
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -82,7 +82,7 @@ There are two ways of running the `goat` locally:
 
 4. Autofix is available for certain linters and is enabled by default. To disable autofix, use:  
    `docker run -e INPUT_AUTO_FIX="false" -v "$PWD:/goat/" --rm <hash>` or  
-   `pipenv run invoke goat --disable-reformat`
+   `pipenv run invoke goat --disable-autofix`
 
 ### Linter Update Considerations
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,11 +80,9 @@ There are two ways of running the `goat` locally:
     Note: Linter env variables must be formatted as <LINTER_CONFIG>, i.e. RUFF_CONFIG, CFN_LINT_CONFIG,
     etc., and the values supplied will overwrite the default arguments supplied in the goat.
 
-4. Autofix is available for certain linters, is disabled by default when running in a pipeline,  
-   and is enabled by default when running locally. To disable autofix locally, use:  
-   `docker run -e INPUT_AUTO_FIX="false" -v $PWD:/goat/ -rm <hash>` or  
-   `pipenv run invoke noreformat`
-   To enable autofix when running in a pipeline, pass `INPUT_AUTO_FIX="true"`. This will result in a dirty git directory.
+4. Autofix is available for certain linters and is enabled by default. To disable autofix, use:  
+   `docker run -e INPUT_AUTO_FIX="false" -v "$PWD:/goat/" --rm <hash>` or  
+   `pipenv run invoke goat --disable-reformat`
 
 ### Linter Update Considerations
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,9 +80,11 @@ There are two ways of running the `goat` locally:
     Note: Linter env variables must be formatted as <LINTER_CONFIG>, i.e. RUFF_CONFIG, CFN_LINT_CONFIG,
     etc., and the values supplied will overwrite the default arguments supplied in the goat.
 
-4. Autofix is available for certain linters and can be invoked in two ways:
-   1. `docker run -e INPUT_AUTO_FIX="true" -v $PWD:/goat/ -rm <hash>`
-   2. `pipenv run invoke reformat`
+4. Autofix is available for certain linters, is disabled by default when running in a pipeline,  
+   and is enabled by default when running locally. To disable autofix locally, use:  
+   `docker run -e INPUT_AUTO_FIX="false" -v $PWD:/goat/ -rm <hash>` or  
+   `pipenv run invoke noreformat`
+   To enable autofix when running in a pipeline, pass `INPUT_AUTO_FIX="true"`. This will result in a dirty git directory.
 
 ### Linter Update Considerations
 
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ jobs:
   
 1. Autofix code formatting errors using those linters with a built-in `fix` option. Note, this is enabled by default  
    when running locally and can create a `dirty` git directory. You will need to manage committing and pushing any  
-   changes. Setting `INPUT_AUTO_FIX=false` locally will disable autofix and setting the action's auto_fix option to  
+   changes.
    `true` will force autofix to run in a pipeline.
 
     ```bash
--- a/README.md
+++ b/README.md
@@ -94,7 +94,6 @@ jobs:
 1. Autofix code formatting errors using those linters with a built-in `fix` option. Note, this is enabled by default  
    when running locally and can create a `dirty` git directory. You will need to manage committing and pushing any  
    changes.
-   `true` will force autofix to run in a pipeline.
 
     ```bash
     uses: seisollc/goat@main
--- a/README.md
+++ b/README.md
@@ -91,13 +91,13 @@ jobs:
     EOF
     ```
   
-1. Autofix code formatting errors using those linters with a built-in `fix` option. Note, this will create a `dirty` git directory,  
-   and you will need to manage committing and pushing any changes; consider setting `INPUT_AUTO_FIX=true` locally instead.
+1. Autofix code formatting errors using those linters with a built-in `fix` option. Note, this is enabled by default when running  
+   locally and will create a `dirty` git directory. You will need to manage committing and pushing any changes. Setting `INPUT_AUTO_FIX=false` locally will disable autofix and setting the action's auto_fix option to `true` will force autofix to run in a pipeline.
 
     ```bash
     uses: seisollc/goat@main
     with:
-      auto_fix: true
+      auto_fix: false
     ```
 
 #### Supported Linters
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ jobs:
     ```
   
 1. Autofix code formatting errors using those linters with a built-in `fix` option. Note, this is enabled by default  
-   when running locally and will create a `dirty` git directory. You will need to manage committing and pushing any  
+   when running locally and can create a `dirty` git directory. You will need to manage committing and pushing any  
    changes. Setting `INPUT_AUTO_FIX=false` locally will disable autofix and setting the action's auto_fix option to  
    `true` will force autofix to run in a pipeline.
 
--- a/README.md
+++ b/README.md
@@ -91,8 +91,10 @@ jobs:
     EOF
     ```
   
-1. Autofix code formatting errors using those linters with a built-in `fix` option. Note, this is enabled by default when running  
-   locally and will create a `dirty` git directory. You will need to manage committing and pushing any changes. Setting `INPUT_AUTO_FIX=false` locally will disable autofix and setting the action's auto_fix option to `true` will force autofix to run in a pipeline.
+1. Autofix code formatting errors using those linters with a built-in `fix` option. Note, this is enabled by default  
+   when running locally and will create a `dirty` git directory. You will need to manage committing and pushing any  
+   changes. Setting `INPUT_AUTO_FIX=false` locally will disable autofix and setting the action's auto_fix option to  
+   `true` will force autofix to run in a pipeline.
 
     ```bash
     uses: seisollc/goat@main
--- a/action.yml
+++ b/action.yml
@@ -15,6 +15,7 @@ inputs:
   auto_fix:
     description: "Auto-fix code formatting errors"
     required: false
+    default: "true"
 runs:
   using: "docker"
   image: "docker://docker.io/seiso/goat:latest"
--- a/action.yml
+++ b/action.yml
@@ -15,7 +15,6 @@ inputs:
   auto_fix:
     description: "Auto-fix code formatting errors"
     required: false
-    default: "false"
 runs:
   using: "docker"
   image: "docker://docker.io/seiso/goat:latest"
--- a/action.yml
+++ b/action.yml
@@ -15,6 +15,7 @@ inputs:
   auto_fix:
     description: "Auto-fix code formatting errors"
     required: false
+    default: "false"
 runs:
   using: "docker"
   image: "docker://docker.io/seiso/goat:latest"
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -55,11 +55,9 @@ function setup_environment() {
 		export FILTER_REGEX_EXCLUDE="${INPUT_EXCLUDE}"
 	fi
 	
-	if [[ ${INPUT_AUTO_FIX:-} == "false" ]]; then
+	if [[ ${INPUT_AUTO_FIX:-true} == "false" ]]; then
 		# Let INPUT_AUTO_FIX override the autofix value. This allows for disabling autofix locally.
 		AUTO_FIX="false"
-	elif [[ ${INPUT_AUTO_FIX:-} == "true" ]]; then
-		LINT_ROUND=2
 	fi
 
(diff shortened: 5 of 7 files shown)


tasks.py
```diff
@@ -252,20 +252,20 @@ def goat(_c, debug=False):
 
 
 @task
-def reformat(_c, debug=False):
-    """Reformat the goat"""
+def noreformat(_c, debug=False):
```
ChiefHolland: I don't like noreformat, but Invoke didn't like hyphens. I'm open to suggestions if this baby is too ugly.

entrypoint.sh
```diff
@@ -89,6 +93,15 @@ ${overlap}"
 		feedback ERROR "The repo dictionary must be sorted"
 		exit 1
 	fi
+
+	if [[ ${CI:-false} == "true" || ${CONTINUOUS_INTEGRATION:-false} == "true" || ${GITHUB_ACTIONS:-false} == "true" ]]; then
+		export IS_PIPELINE="true"
+		# If this is run in a pipeline and INPUT_AUTO_FIX does not equal true, disable autofix. INPUT_AUTO_FIX="true" would force autofix
```
ChiefHolland: This is an assumption on my part that any changes made by an action on a repo get discarded when the run completes. I also assume we can check for a dirty repo in the pipeline. Is this correct?

entrypoint.sh
```diff
@@ -304,20 +312,48 @@ runtime=$((end - start))
 echo -e "\nScanned ${#included[@]} files in ${runtime} seconds"
 echo -e "Excluded ${#excluded[@]} files\n"
 
-for success in "${linter_successes[@]}"; do
-	feedback INFO "$success completed successfully"
-done
+linting_failed="false"
```
ChiefHolland: I added this to offer a failed status when autofix is enabled in a pipeline. The linters return 0 when fix is successful.

entrypoint.sh
JonZeolla: That's true, unless you explicitly add, commit, and push them (which you can do, and is how our autoupdate logic works) it is discarded at the end of the job.

JonZeolla: Why can't we just rely on the `CI` env var?  In practice I have always seen it properly set in a pipeline, and then we wouldn't even need the IS_PIPELINE, we can just check `CI` directly unless there are edge cases I don't know about

entrypoint.sh
JonZeolla: It shouldn't fail if autofix is enabled in a pipeline because a future step may choose to add, commit, and push the dirty git directory.

entrypoint.sh
ChiefHolland: That was a ChatGPT-ism. It gave both as options, but if there isn't a significant use case for other than CI, I'm totally game for simplifying it.

CONTRIBUTING.md
```diff
@@ -80,9 +80,11 @@ There are two ways of running the `goat` locally:
     Note: Linter env variables must be formatted as <LINTER_CONFIG>, i.e. RUFF_CONFIG, CFN_LINT_CONFIG,
     etc., and the values supplied will overwrite the default arguments supplied in the goat.
 
-4. Autofix is available for certain linters and can be invoked in two ways:
-   1. `docker run -e INPUT_AUTO_FIX="true" -v $PWD:/goat/ -rm <hash>`
-   2. `pipenv run invoke reformat`
+4. Autofix is available for certain linters, is disabled by default when running in a pipeline,
```
JonZeolla: I thought for a pipeline we wanted to have a default that:
1. Runs the linters.
2. If any of the linters failed and auto fix was enabled (the default), it runs autofix.
3. Re-run linting.  If it now succeeds, (1) print the linter errors, and after those (2) print a message indicating that the issues _could_ be autofixed using <insert the command the user should run to autofix, like `pipenv run invoke reformat`; exit non-zero.  If the second linting still fails, just print the linter errors and exit non-zero

entrypoint.sh
ChiefHolland: Cool. That means I can get rid of this superfluous var.

CONTRIBUTING.md
```diff
@@ -80,9 +80,11 @@ There are two ways of running the `goat` locally:
     Note: Linter env variables must be formatted as <LINTER_CONFIG>, i.e. RUFF_CONFIG, CFN_LINT_CONFIG,
     etc., and the values supplied will overwrite the default arguments supplied in the goat.
 
-4. Autofix is available for certain linters and can be invoked in two ways:
-   1. `docker run -e INPUT_AUTO_FIX="true" -v $PWD:/goat/ -rm <hash>`
-   2. `pipenv run invoke reformat`
+4. Autofix is available for certain linters, is disabled by default when running in a pipeline,  
+   and is enabled by default when running locally. To disable autofix locally, use:  
+   `docker run -e INPUT_AUTO_FIX="false" -v $PWD:/goat/ -rm <hash>` or
```
```suggestion
   `docker run -e INPUT_AUTO_FIX="false" -v "$PWD:/goat/" --rm <hash>` or  
```

tasks.py
JonZeolla: @ChiefHolland if you wanted hyphens you just do something like:

```suggestion
def no_reformat(_c, debug=False):
```

and run `pipenv run invoke no-reformat`

JonZeolla: That said, I don't really understand the "no reformat" thing, aren't we just running the goat / standard linters?  You can update the arguments as well, remember this is just python.  So you could set a `reformat=True` argument or something.  [Here's an example](https://github.com/SeisoLLC/easy_infra/blob/4d216d4f0db4d649f1371eeeb80ebb04ef22331b/tasks.py#L644)

JonZeolla: Also from [the docs](https://www.pyinvoke.org/):

```
Following the lead of most Unix CLI applications, it offers a traditional flag-based style of command-line parsing, deriving flag names and value types from task signatures (optionally, of course!):

$ invoke clean --docs --bytecode build --docs --extra='**/*.pyo'
$ invoke clean -d -b build --docs -e '**/*.pyo'
$ invoke clean -db build -de '**/*.pyo'
````

File: entrypoint.sh [hunk 6]
```diff
@@ -304,20 +312,48 @@ runtime=$((end - start))
...
+		fi
+	fi
+
+	for success in "${linter_successes[@]}"; do
+		feedback INFO "$success completed successfully"
+	done
+fi
 
-if [ -n "${linter_failures[*]}" ]; then
+if [ -n "${linter_failures[*]}" ]; then
```
```suggestion
if [ -n "${linter_failures[*]}" ]; then
```

JonZeolla (changes requested):
Didn't do a full review because I think some of my feedback may result in some bigger changes

CONTRIBUTING.md
JonZeolla: From my Wrike task comment:
```
1. If you are in a pipline (CI is true) then print a message saying "This can be autofixed!"
2. If you are not in a pipeline (CI is not true) then say "We autofixed your $linter failures, check it out and commit if you agree"
```

JonZeolla: Does this all make sense?

To be more explicit, locally I was thinking the default would be:
1. Run the linters
2. If any linters failed and autofix is enabled (default), run autofix.  Make sure to print a warning: Saw issues, trying to autofix...
3. Re-run linting and if successful, print the "we autofixed your $linter failures, check it out and commit" message.  Exit non-zero because the user needs to take action (review logs and commit).
4. If the second linter fails, just print the logs and err out stating we tried to autofix but it wasn't successful; your local dir may be dirty with some partial fixes.

tasks.py
ChiefHolland: Yanked it and added a parameter to the invoke goat function to turn off autofix.

entrypoint.sh
ChiefHolland: Removed this and opted for a local check for CI.

entrypoint.sh
```diff
@@ -46,13 +50,18 @@ function setup_environment() {
...
 	if [[ -n ${INPUT_EXCLUDE:+x} ]]; then
 		export FILTER_REGEX_EXCLUDE="${INPUT_EXCLUDE}"
 	fi
+	
+	if [[ ${INPUT_AUTO_FIX:-} == "false" ]]; then
+		# Let INPUT_AUTO_FIX override the autofix value. This allows for disabling autofix locally.
+		AUTO_FIX="false"
+	fi
+
+	if [[ ${INPUT_AUTO_FIX:-} == "true" ]]; then
```
ChiefHolland: I added this to have more control on forcing autofix on or off.

entrypoint.sh
```diff
@@ -159,6 +169,34 @@ function get_files_matching_filetype() {
 	echo "${matching_files[@]}"
 }
 
+function load_filetype_array() {
```
ChiefHolland, line 168: Since there was more in seiso_lint than just linting, I needed a way to recreate the linter objects without duplicating a lot of code. This is a start. I plan on cleaning up and making functions of a few more aspects in future sprints.

entrypoint.sh
```diff
@@ -286,14 +312,46 @@ function seiso_lint() {
...
 		else
-			if [[ ${AUTO_FIX:-} == "true" ]]; then
-				cat "/opt/goat/log/${pids[$p]}.log"
-			fi
 			linter_successes+=("${pids[$p]}")
 		fi
 	done
 }
 
+function rerun_lint() {
```
ChiefHolland, line 316: Once I strip out the lint functions to their core functionality and learn to pass variables back and forth a little better, we'll have a single lint function.

entrypoint.sh
ChiefHolland: Or not :D

CONTRIBUTING.md
ChiefHolland: I think I have all of the options captured.

entrypoint.sh
JonZeolla: I don't understand the changes here as a whole.  Couldn't it be a single if/else?  Also can you explain what `RE_RUN` is a bit more.  Are we saying that this is a re_run if it's true, or are we saying it turns on the ability to re_run if it's true?

.github/etc/dictionary.txt
```diff
@@ -3,4 +3,5 @@ capricornis
 crispus
 hadolint
 kubeconform
+noreformat
```
JonZeolla: No longer needed
```suggestion
```

README.md
```diff
@@ -91,13 +91,15 @@ jobs:
     EOF
     ```
   
-1. Autofix code formatting errors using those linters with a built-in `fix` option. Note, this will create a `dirty` git directory,  
-   and you will need to manage committing and pushing any changes; consider setting `INPUT_AUTO_FIX=true` locally instead.
+1. Autofix code formatting errors using those linters with a built-in `fix` option. Note, this is enabled by default  
+   when running locally and will create a `dirty` git directory. You will need to manage committing and pushing any
```
```suggestion
   when running locally and can create a `dirty` git directory. You will need to manage committing and pushing any  
```

README.md
```diff
@@ -91,13 +91,15 @@ jobs:
     EOF
     ```
   
-1. Autofix code formatting errors using those linters with a built-in `fix` option. Note, this will create a `dirty` git directory,  
-   and you will need to manage committing and pushing any changes; consider setting `INPUT_AUTO_FIX=true` locally instead.
+1. Autofix code formatting errors using those linters with a built-in `fix` option. Note, this is enabled by default  
+   when running locally and will create a `dirty` git directory. You will need to manage committing and pushing any  
+   changes. Setting `INPUT_AUTO_FIX=false` locally will disable autofix and setting the action's auto_fix option to
```
JonZeolla: No need for the extra explainer in README.md, it is covered in CONTRIBUTING.  Normal use case doesn't care about running locally. 
```suggestion
   changes.
```

README.md
```diff
@@ -91,13 +91,15 @@ jobs:
     EOF
     ```
   
-1. Autofix code formatting errors using those linters with a built-in `fix` option. Note, this will create a `dirty` git directory,  
-   and you will need to manage committing and pushing any changes; consider setting `INPUT_AUTO_FIX=true` locally instead.
+1. Autofix code formatting errors using those linters with a built-in `fix` option. Note, this is enabled by default  
+   when running locally and will create a `dirty` git directory. You will need to manage committing and pushing any  
+   changes. Setting `INPUT_AUTO_FIX=false` locally will disable autofix and setting the action's auto_fix option to  
+   `true` will force autofix to run in a pipeline.
```
```suggestion
```

action.yml
```diff
@@ -15,6 +15,7 @@ inputs:
   auto_fix:
     description: "Auto-fix code formatting errors"
     required: false
+    default: "false"
```
JonZeolla: Shouldn't we turn it on by default in pipelines too?  That will result in the better error "we could have fixed this" right?

entrypoint.sh
```diff
@@ -62,6 +71,7 @@ function setup_environment() {
 	declare -a linter_failures
 	declare -a linter_successes
 	declare -a linter_skipped
+	input="/etc/opt/goat/linters.json"
```
JonZeolla: Seems like a strange place to put this.  Why do we need it here?  I would think we just export a global `INPUT` and rename it `LINTER_CONFIG`

entrypoint.sh
JonZeolla: Sounds good

entrypoint.sh
```diff
@@ -159,6 +169,34 @@ function get_files_matching_filetype() {
...
+}
+
+function has_autofix() {
+	local lint_name=$1
+
+	while IFS= read -r line; do
+		name=$(echo "$line" | jq -r ".name")
+		if [[ "$name" == "$lint_name" ]]; then
+			if echo "$line" | jq -e '.autofix' > /dev/null; then
+				# If linter has an autofix exit successfully
```
JonZeolla: We're mixing how we use return.  In other functions return 0 means we didn't find one like [here](https://github.com/SeisoLLC/goat/blob/a967dfc01eeabf4a3cfea5a611fc448d1b353276/entrypoint.sh#L94-L108).  Whatever we choose, let's stay consistent

entrypoint.sh
```diff
@@ -159,6 +169,34 @@ function get_files_matching_filetype() {
...
+				return 0
+			else
+				return 1
+			fi
+		fi
+	done < <(jq -c '.[]' $input)
```
JonZeolla: Need to quote this in case it has spaces.  Also, it makes linters happy
```suggestion
	done < <(jq -c '.[]' "${input}")
```

entrypoint.sh
```diff
@@ -167,6 +205,10 @@ function lint_files() {
 	local files_to_lint=""
 	local env_var_name="${linter_array[env]}"
 
+	if [ ${RE_RUN:-false} == "true" ] && has_autofix "${linter_array[name]}"; then
```
JonZeolla: Based on this, I think `RE_RUN` means "this is a re_run" but it's still a little bit of a guess for me.  I think what would be clearer is calling the var something like `LINT_ROUND` or `ITERATION` and set it to 1 for the first time and 2 for the re run.  Or something like that so it's readable/understandable

entrypoint.sh
```diff
@@ -286,14 +312,46 @@ function seiso_lint() {
...
-			if [[ ${AUTO_FIX:-} == "true" ]]; then
-				cat "/opt/goat/log/${pids[$p]}.log"
-			fi
 			linter_successes+=("${pids[$p]}")
 		fi
 	done
 }
 
+function rerun_lint() {
+	local failed_linter=$1
```
```suggestion
	local failed_linter="$1"
```

entrypoint.sh
```diff
@@ -286,14 +312,46 @@ function seiso_lint() {
...
+					rerun_filetypes=("$(load_filetype_array "$value")")
+					continue
+				fi
+				value=$(echo "$value" | tr -d "'" | tr -d '"')
+				rerun_linter["$key"]=$value
+			done < <(echo "$line" | jq -r 'to_entries|map("\(.key)=\(.value|tojson)")|.[]')
+
+			rerun_linter[logfile]="/opt/goat/log/rerun_${rerun_linter[name]}.log"
+		fi
+	done < <(jq -c '.[]' $input)
```
```suggestion
	done < <(jq -c '.[]' "${input}")
```

JonZeolla (changes requested):
Some really quick feedback

entrypoint.sh
ChiefHolland: Autofix is enabled by default, both in the pipeline and locally. The INPUT_AUTO_FIX env has two purposes for greater control over the functionality: 1) true bypasses the initial scan and selects the autofix arg set for those linters that have them and the standard args for those that don't, and 2) false disables autofix and only runs the initial scan without reformatting. The RE_RUN variable is for flow control in selecting the correct args to use.

action.yml
ChiefHolland: Autofix is enabled by default. I meant to remove this line after my latest refactor. Defaulting this variable as either true or false changes the natural flow of the goat. Default is scan once for initial and twice if errors that can be autofixed. Set auto_fix=true to bypass the initial scan and only autofix. Set auto_fix=false to only run an initial scan with no autofix.

entrypoint.sh
ChiefHolland: Oh, I like that better. As to why I put it there, I'm used to lumping global vars or static values at the top of an app. Carryover from C#, I'm sure. I'll add this change, though. It flows better from a bash perspective.

entrypoint.sh
ChiefHolland: Will do

JonZeolla (approved):
🚀

PR #111, opened 2023-06-26
fix: pre-commit permissions

JonZeolla:
# Contributor Comments

pre-commit specifies a uid/gid if it's available, breaking some logging permissions (see [here](https://github.com/pre-commit/pre-commit/blob/c25ea47cf78b42da8fd403c435f156b056aaeb38/pre_commit/languages/docker.py#L106))

This was a little tricky because the `pipenv run` actually runs two separate installs: (1) pipenv virtual environment (expected), and (2) a normal pip install (which was installing following https://peps.python.org/pep-0370/) which was unexpected.  Had to add support for both by any arbitrary user in the goat for this to work.

## Testing

Please check that you can install and run pre commit locally. It should pass

## Pull Request Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,8 +48,10 @@ RUN pip install pipenv \
     jscpd \
     markdown-link-check \
     && mkdir -p /opt/goat/log \
-    # Necessary because pre-commit adds -u os.uid():os.gid() to the docker run
-    && chmod o+w /opt/goat/log
+    # The following commands are necessary because pre-commit adds -u os.uid():os.gid() to the docker run
+    && chmod o+w /opt/goat/log \
+    && mkdir -p /.local \
+    && chmod o+w /.local
 
 WORKDIR /goat/
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,7 +47,9 @@ RUN pip install pipenv \
     cspell \
     jscpd \
     markdown-link-check \
-    && mkdir -p /opt/goat/log
+    && mkdir -p /opt/goat/log \
+    # Necessary because pre-commit adds -u os.uid():os.gid() to the docker run
+    && chmod o+w /opt/goat/log
 
 WORKDIR /goat/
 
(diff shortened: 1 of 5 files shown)


ChiefHolland (dismissed):
Blech. But okay.

Dockerfile
```diff
@@ -21,7 +21,7 @@ LABEL org.opencontainers.image.licenses="MIT"
 
 WORKDIR /etc/opt/goat/
 ENV PIP_NO_CACHE_DIR=1
-ENV PIPENV_PIPFILE="/etc/opt/goat/Pipfile"
+ENV WORKON_HOME=/tmp
```
JonZeolla, line 24: This puts the venv in /tmp to work around perms issues.

Dockerfile
```diff
@@ -47,12 +47,14 @@ RUN pip install pipenv \
     cspell \
     jscpd \
     markdown-link-check \
-    && mkdir -p /opt/goat/log
+    && mkdir -p /opt/goat/log \
+    # The following commands are necessary because pre-commit adds -u os.uid():os.gid() to the docker run
+    && chmod o+w /opt/goat/log \
+    && mkdir -p /.local \
```
JonZeolla, line 53: This fixes the following error:
```bash
pipenv run mypy --config-file /etc/opt/goat/.mypy.ini --install-types --non-interactive ./tasks.py
Defaulting to user installation because normal site-packages is not writeable
Collecting types-requests
  Downloading types_requests-2.31.0.1-py3-none-any.whl (14 kB)
Collecting types-urllib3
  Downloading types_urllib3-1.26.25.13-py3-none-any.whl (15 kB)
Installing collected packages: types-urllib3, types-requests
ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied: '/.local'
Check the permissions.
```

JonZeolla: Specifically note this line:
Defaulting to user installation because normal site-packages is not writeable

ChiefHolland (approved):
Glad you figured it out! Looks good.

PR #118, opened 2023-07-05
Fix pyenv

JonZeolla:
# Contributor Comments

This fixes the `pyenv` install/environment setup so that `pipenv` can swap to arbitrary versions of `python` as needed in order to perform linting

## Pull Request Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ LABEL org.opencontainers.image.licenses="MIT"
 WORKDIR /etc/opt/goat/
 ENV PIP_NO_CACHE_DIR=1
 ENV WORKON_HOME=/tmp
-ENV PYENV_ROOT="$HOME/.pyenv"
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -47,16 +49,16 @@ RUN pip install pipenv \
     cspell \
     jscpd \
     markdown-link-check \
+    && git clone https://github.com/pyenv/pyenv.git --depth=1 "${PYENV_ROOT}" \
+    && find "${PYENV_ROOT}" -type d -name ".git" -exec rm -rf {} + \
+    && eval "$(pyenv init -)" \
```
JonZeolla: Need do to this before the find to cleanup the plugins

JonZeolla: Not true; plugins don't come in via `init` 🤔

ChiefHolland: I think I remember reading that the init loads the plug-in paths to $PATH.

PR #120, opened 2023-07-11
fix: add tk-dev for pyenv

ChiefHolland:
# Contributor Comments

An error was discovered that would prevent pipenv from creating a virtual environment due to pyenv not being able to complete installing the necessary version of Python. Pyenv's recommended dependencies are listed [here](https://github.com/pyenv/pyenv/wiki#suggested-build-environment). The minimum viable addition to the goat for full functionality is tk-dev, which was added to the Dockerfile.

## Pull Request Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,8 @@ RUN pip install pipenv \
     tini \
     bash \
     git \
+    # The following apk package is necessary for pyenv functionality
+    tk-dev \
     && npm install --save-dev --no-cache -g dockerfile_lint \
     markdownlint-cli \
     textlint \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -50,8 +52,8 @@ RUN pip install pipenv \
     jscpd \
     markdown-link-check \
     && git clone https://github.com/pyenv/pyenv.git --depth=1 "${PYENV_ROOT}" \
-    && echo 'command -v pyenv >/dev/null || export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.profile \
-    && echo 'eval "$(pyenv init -)"' >> ~/.profile \
+    && echo 'command -v pyenv >/dev/null || export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.bash_profile \
```
ChiefHolland, line 55: https://github.com/pyenv/pyenv/wiki#which-shell-startup-file-do-i-put-pyenv-config-in

Dockerfile
```diff
@@ -40,6 +40,8 @@ RUN pip install pipenv \
     tini \
     bash \
     git \
+    # The following apk package is necessary for pyenv functionality
+    tk-dev \
```
ChiefHolland, line 44: https://github.com/pyenv/pyenv/wiki#suggested-build-environment

ChiefHolland: tk-dev was the single package blocking our current use case. Logging here in case we need to add more in the future.

Dockerfile
ChiefHolland: Did we determine we were using bash or ash? If ash, we can revert back to .profile.

JonZeolla: nit: bash is good, but isn't `.profile` just a more generic version of `.bash_profile`?  We have no reason to be more specific

JonZeolla (approved):

Dockerfile
ChiefHolland: I agree we don't have a need for more specificity. My suggestion was more to be consistent with the documentation I was using for justifying tk-dev.

PR #123, opened 2023-07-17
feat: Use task lint for pre-commit

JonZeolla:
# Contributor Comments

This is so we can align projects that use this pre-commit hook with their own `task lint` command. Then, we can set vars in the user's `Taskfile.yml` and assess that automatically, while avoiding the use of `env`s which can be different in a local vs pipeline, etc.

It also simplifies the use of `goat` configurations like exemptions; you can set that in your `Taskfile.yml` instead of trying to using the `pre-commit` [`exclude`](https://pre-commit.com/#top_level-exclude) feature which can be divergent from the goat's [`exclude`](https://github.com/SeisoLLC/goat#customizations) feature, and confusing on which to pick.

## Pull Request Checklist


Task/python/Taskfile.yml
```diff
@@ -27,8 +27,19 @@ tasks:
   lint:
     preconditions:
       - sh: docker pull seiso/goat:latest || echo "Failed to pull the latest goat, continuing anyway..."
+    # Adjustments are passed as vars, not envs, to avoid local workarounds
```
ChiefHolland, line 30: <3

entrypoint.sh
```diff
@@ -63,6 +63,18 @@ function setup_environment() {
 
 	export REPO_DICTIONARY="${RELATIVE_PATH}/.github/etc/dictionary.txt"
 
+	#############
```
ChiefHolland, line 66: Thanks for this reminder!

seanctech (dismissed):
LGTM

seanctech (approved):
LGTM

ChiefHolland (approved):
LGTM

PR #129, opened 2023-07-18
fix: automaintain pre-commit releases

JonZeolla:
# Contributor Comments

automaintain pre-commit releases.  In 754b9ec5f2e08e199a4ff043ceb363ab27a0c019 I froze the pre-commit hash after a failed pipeline causing the hash in the repo to already be fully up to date, meaning that the `task pre-commit-update -- commit` in the pipeline had nothing to do, causing it to fail.  In order to fix this, we need to push an older pre-commit hash, so the update will have something to do.

Give those 🤖 some work!

## Pull Request Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 ---
 repos:
   - repo: https://github.com/seisollc/goat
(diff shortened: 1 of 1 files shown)


ChiefHolland (approved):
LGTM

PR #130, opened 2023-07-18
fix: don't pass filenames to pre-commit

JonZeolla:
# Contributor Comments

Don't pass filenames to pre-commit (docs [here](https://pre-commit.com/#hooks-pass_filenames)); by default it will take your `entry` and pass the detected filenames to it. In our case, that detection happens inside of the `task` job (which `docker run`s the goat), so we need to turn that off.

If you hit this you'll get all kinds of strange issues like:

```bash
task: No tasks with description available. Try --list-all to list all tasks
task: Task ".pre-commit-config.yaml" does not exist
```

## Pull Request Checklist

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -3,3 +3,4 @@
   name: Lint (via the 🐐)
   language: system
   entry: task lint
+  pass_filenames: false


ChiefHolland (approved):
Lgtm

PR #131, opened 2023-07-24
refactor: move to taskfile

JonZeolla:
# Contributor Comments

This is a refactor to move from invoke to taskfiles.

❗️This may not work on initial commit because the submodule will need to be updated to point to the latest release of the goat.

**Important notes**

1. Note that this removes the `run_security_tests` which used to run trivy but would exit 0 regardless and wasn't really operationally reviewed. We can re-add it in if we are going to actually leverage the security scanning information (consider syft/grype instead, etc.)
2. There is a lot of refactoring of how vars are passed around, which supports more concise variable usage.
3. This now leverages the "base" Taskfile.yml much more, and has adjustments for bash vs python. This was an expected follow-on to the initial python/Taskfile.yml.
4. Instead of replacing `pipenv run invoke goat` with `task goat` we are now running `task test`, which is more standard.
5. Please test by pointing a Python project’s submodule to this branch

## Testing

This is a significant change and should be manually tested. Please:

1. Clone this branch, follow the `README.md` instructions, and run some `task` commands. I especially recommend simulating the GitHub Actions as those are more difficult to test.
2. Generate a paved road project, update the submodule to point to this branch, pull the changes down, and run the base commands to ensure it works. However, this will not go into effect until downstream projects update their submodule, which is by design so if there's a bug it won't break all of our builds 😀

## Pull Request Checklist


.github/CONTRIBUTING.md
```diff
@@ -86,9 +86,9 @@ There are two ways of running the `goat` locally:
     and the values supplied will take precedence over the default autofix or standard arguments supplied in the goat.
     Any desired autofix arguments must be explicitly supplied as part of the new env variable value.
 
-4. Autofix is available for certain linters and is enabled by default. To disable autofix, use:
+4. Autofix is available for certain linters and is enabled by default. To disable autofix, use one of these approaches:
    1. `docker run -e INPUT_AUTO_FIX="false" -v "$PWD:/goat/" --rm <hash>` or
-   2. `pipenv run invoke goat --disable-autofix`
+   2. `INPUT_AUTO_FIX=false task test`
```
JonZeolla, line 91: Need to test

.github/workflows/commit.yml
```diff
@@ -30,13 +32,15 @@ jobs:
...
             ${{ runner.os }}-python-
       - name: Install the dependencies
-        run: |
-          python -m pip install --upgrade pipenv
-          pipenv install --python ${{ env.python_version }} --deploy --ignore-pipfile --dev
+        run: python -m pip install --upgrade pipenv
+      - name: Install Task
+        uses: arduino/setup-task@v1
+      - name: Initialize the repo
+        run: task init
```
JonZeolla: We may need to update clone instructions to include `--submodules`

Task/Taskfile.yml
```diff
@@ -4,6 +4,107 @@
 version: "3"
 
 tasks:
+  lint:
+    preconditions:
+      - sh: docker pull seiso/goat:latest || echo "Failed to pull the latest goat, continuing anyway..."
+    # Adjustments are passed as vars, not envs, to avoid local workarounds
+    vars:
```
JonZeolla, line 41: This is new; required so we can pass the vars around more directly instead of doing global vars

Task/Taskfile.yml
```diff
@@ -4,6 +4,107 @@
...
+      - |
+        docker run --env INPUT_AUTO_FIX="{{.INPUT_AUTO_FIX}}" \
+                   --env INPUT_DISABLE_MYPY="{{.INPUT_DISABLE_MYPY}}" \
+                   --env INPUT_EXCLUDE="{{.INPUT_EXCLUDE}}" \
+                   --env INPUT_LOG_LEVEL="{{.INPUT_LOG_LEVEL}}" \
+                   --rm -v "{{.ROOT_DIR}}:/goat" seiso/goat:latest
+
+  build:
+    vars:
+      VERSION: '{{.VERSION | default "UNKNOWN"}}'
```
JonZeolla, line 68: This is new; uses the `default` function and will drop to "UNKNOWN" but otherwise takes the provided `VERSION` and uses it.  Downstream (`bash/Taskfile.yml` and `python/`) should be passing a version with logic specific to bash/python.

Task/Taskfile.yml
```diff
@@ -4,6 +4,107 @@
...
+          {{range $tag := .REPO_TAGS | splitLines -}}
+          repo_tags.append("{{$tag}}")
+          {{end}}
+          if (version_string in repo_tags and "{{.TAG_COMMIT_HASH}}" == "{{.COMMIT_HASH}}"):
+              build_version = "{{.VERSION}}"
+          else:
+              build_version = f"{{.VERSION}}-{{.COMMIT_HASH_SHORT}}"
+          print(build_version)'
+    cmds:
+      - docker buildx build --load --build-arg VERSION="{{.BUILD_VERSION}}" --build-arg COMMIT_HASH="{{.COMMIT_HASH}}" --tag {{.IMAGE_NAME}}:latest --tag {{.IMAGE_NAME}}:{{.BUILD_VERSION}} {{.ROOT_DIR}}/.
```
JonZeolla: I added the `{{.ROOT_DIR}}` part of this so it has the right context

Task/Taskfile.yml
```diff
@@ -4,6 +4,107 @@
...
+          else:
+              increment = "01"
+          new_version = f"{{.DATE_INFO}}.{increment}"
+          print(new_version)'
+    cmds:
+      # Better conditional syntax is pending https://github.com/go-task/task/issues/608
+      - '{{if eq .VERSIONING "CalVer"}}pipenv run bumpversion --config-file setup.cfg --new-version {{.RELEASE_VERSION}} \
+        unusedpart{{else if eq .VERSIONING "SemVer"}}pipenv run bumpversion --config-file setup.cfg {{.RELEASE_TYPE}}{{end}}'
+
+  publish:
```
JonZeolla, line 148: This is a brand new `task` primarily to fix semver releases or enable latest-only releases for a subset of calver projects.

.github/CONTRIBUTING.md
```diff
@@ -6,7 +6,7 @@ Ensure you have `docker` and `pipenv` installed locally, and the `docker` daemon
 your local system:
```
JonZeolla, line 6: Don't forget to repoint the submodule to `main`

JonZeolla: I am suggesting we purposefully don't do this so that this merge succeeds, and then after CI/CD completes we repoint it back to `main`.

Dockerfile
```diff
@@ -1,17 +1,23 @@
-FROM ghcr.io/yannh/kubeconform:v0.6.3 as kubeconform
-FROM hadolint/hadolint:v2.12.0-alpine as hadolint
-FROM koalaman/shellcheck:v0.9.0 as shellcheck
-FROM rhysd/actionlint:1.6.25 as actionlint
+# syntax=docker/dockerfile:1
```
JonZeolla, line 1: The `platform` work here and in the `Taskfile.yml` files was necessary for my local testing/confirmation that the goat works.

Task/Taskfile.yml
```diff
@@ -3,8 +3,127 @@
 
 version: "3"
 
+set:
```
JonZeolla, line 6: This essentially runs `set -euo pipefail` which we always want.  Same with the other `Taskfile.yml`s

Task/Taskfile.yml
```diff
@@ -3,8 +3,127 @@
...
+    vars:
+      INPUT_AUTO_FIX: '{{.INPUT_AUTO_FIX | default ""}}'
+      INPUT_DISABLE_MYPY: '{{.INPUT_DISABLE_MYPY | default ""}}'
+      INPUT_EXCLUDE: '{{.INPUT_EXCLUDE | default ""}}'
+      INPUT_LOG_LEVEL: '{{.INPUT_LOG_LEVEL | default ""}}'
+      PLATFORM: '{{.PLATFORM | default ""}}'
+    preconditions:
+      # Temporary workaround until we have an arm64 image on docker hub
+      - sh: |
+          if [[ "$(uname -m)" -ne "arm64" ]]; then
```
JonZeolla, line 50: Once we build **and push** arm64 images, then we can remove this.

JonZeolla: I self-assigned a task for next sprint - https://www.wrike.com/open.htm?id=1167647127

.github/CONTRIBUTING.md
JonZeolla: I have a task for next sprint to fix this - https://www.wrike.com/open.htm?id=1167647370

It is blocked on another task - https://www.wrike.com/open.htm?id=1113530084 - to update the pipelines to build and push multiplatform images

Dockerfile
JonZeolla: See more info in https://docs.docker.com/build/building/multi-platform/#building-multi-platform-images and https://taskfile.dev/usage/#platform-specific-tasks-and-commands

README.md
```diff
@@ -121,7 +121,7 @@ jobs:
 
 #### Debugging
 
-To debug an issue with the goat, configure the log level to either `ERROR`, `WARN`, `NOTICE`, `VERBOSE`, `DEBUG`, or `TRACE`.
+To debug an issue with the goat, configure the log level to either `ERROR`, `WARN`, `INFO`, or `DEBUG`.
```
JonZeolla, line 124: I cleaned this up to only the log levels we actually use and added debug logging because I needed it to troubleshoot some changes that I was making.

entrypoint.sh
```diff
@@ -288,11 +314,14 @@ function seiso_lint() {
...
 		if [[ -n ${FILTER_REGEX_EXCLUDE:+x} && "${file}" =~ ${FILTER_REGEX_EXCLUDE} ]]; then
+			feedback DEBUG "${file} matched the exclusion regex of ${FILTER_REGEX_EXCLUDE}"
 			excluded+=("${file}")
 			continue
+		else
+			feedback DEBUG "${file} didn't match the exclusion regex of ${FILTER_REGEX_EXCLUDE:-not set}"
 		fi
 		included+=("${file}")
-	done < <(find . \( -path "./.git" -prune \) -o \( -type f -print \))
+	done < <(find /goat \( -path "./.git" -prune \) -o \( -type f -print \))
```
JonZeolla: This is needed so we can do the more explicit `^/{{.PROJECT_SLUG}}/goat/.*` exclusion

VERSION
```diff
@@ -0,0 +1 @@
+2023.07.01
```
JonZeolla, line 1: This is a new required file for bash-in-docker projects for managing versions

JonZeolla:
The build failure is just a bootstrapping issue, I can push a tag later

JonZeolla:
CodeQL just needs a config change too because of removing tasks.py

JonZeolla:
May need to add some bumpversion logic as well (run `task release` in the pipeline?)

.github/workflows/commit.yml
JonZeolla: I don’t think we need to

JonZeolla:
Ok the last failures have to do with `jscpd` not properly accounting for the `INPUT_EXCLUDE` variable.  Any thoughts @ChiefHolland ?

One idea is we could append the `INPUT_EXCLUDE` to the `JSCPD_CONFIG` but they use different regex syntax 🤔

ChiefHolland:
> Ok the last failures have to do with `jscpd` not properly accounting for the `INPUT_EXCLUDE` variable.  Any thoughts @ChiefHolland ?
> 
> 
> 
> One idea is we could append the `INPUT_EXCLUDE` to the `JSCPD_CONFIG` but they use different regex syntax 🤔

Most likely due to jscpd file type in linters.json being set to "all". I'll double check the logic for that case.

.github/CONTRIBUTING.md
JonZeolla: I tested and this works

.github/workflows/commit.yml
JonZeolla: We don't; confirmed this works fine with just `Taskfile.yml` updates

Task/Taskfile.yml
JonZeolla: However, note that we do use global vars when they make sense to reduce duplication or excessive passing around of vars

.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "goat"]
 	path = goat
-	url = https://github.com/SeisoLLC/goat
-	branch = main
+	url = https://github.com/seisollc/goat
+	branch = refactor/taskfiles
```
JonZeolla, line 4: We need to undo this after this merges; there's a follow-on task for that

.github/CONTRIBUTING.md
```diff
@@ -18,7 +18,7 @@ There are two ways of running the `goat` locally:
 1. To see how it will look in a pipeline:
 
     ```bash
-    pipenv run invoke goat
+    task test
     ```
 
     Or add `--debug` for more information.
```
ChiefHolland:
```suggestion
    To see more information, add debug messages to output with `task test -- debug`  or enable verbose mode `task -v test`.
```

README.md
```diff
@@ -1,7 +1,7 @@
 # Grand Opinionated AutoTester (GOAT)
 
-The Grand Opinionated AutoTester (GOAT) automatically applies Seiso's standard testing.
-This project has been designed to be easier to understand and more opinionated than other open source multi-linter projects.
+The Grand Opinionated AutoTester (GOAT) automatically applies Seiso's policy as code. This project is designed to be easier to understand and more opinionated
+than other open source multi-linter projects, and easily extended to programmatically enforce Seiso policies.
```
ChiefHolland:
```suggestion
than other open source multi-linter projects, and easily extended to enforce Seiso policies programmatically.
```

ChiefHolland: nit: split infinitive

ChiefHolland (commented):
Feedback in comments.

.github/CONTRIBUTING.md
JonZeolla: Ah that was just at typo let me fix it.

README.md
JonZeolla: Done!

.github/CONTRIBUTING.md
```diff
@@ -21,7 +21,7 @@ There are two ways of running the `goat` locally:
     task test
     ```
 
-    Or add `--debug` for more information.
+    Or run `task test -- debug` run it in debug mode.
```
ChiefHolland:
```suggestion
    Or run `task test -- debug` to run it in debug mode.
```

ChiefHolland (approved):
LGTM

PR #141, opened 2023-08-04
ci: avoid pre-commit in pipelines

JonZeolla:
# Contributor Comments

This avoids pre-commit installation in pipelines

## Pull Request Checklist

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -89,8 +89,6 @@ jobs:
         uses: arduino/setup-task@v1
       - name: Initialize the repo
         run: task -v init
-      - name: Build the goat
-        run: task -v build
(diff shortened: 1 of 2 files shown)


.github/workflows/commit.yml
```diff
@@ -89,8 +89,6 @@ jobs:
         uses: arduino/setup-task@v1
       - name: Initialize the repo
         run: task -v init
-      - name: Build the goat
-        run: task -v build
```
JonZeolla, line 93: We only did this before so that the local goat that was available for the release pre-commit hook was up-to-date.  We should avoid that overhead altogether when just releasing; the commit it creates will be linted, and this was linted already.

ChiefHolland (approved):
LGTM

PR #150, opened 2023-08-07
refactor: replace goat submodule with a symlink

JonZeolla:
# Contributor Comments

We've been having issues with the goat-in-goat submodule due to recursive cloning in upstream tools. I expect this issue to continue.

Previously I tried to avoid using a goat-in-goat submodule because it is messy, but because the Task `dir:` paths must be relative, I didn't have a great alternative at the time.

Today I came up with the idea to use a symlink instead of a submodule. This is nice because it removes the need to autoupdate the submodule 🥰, and it maintains the same relative path structure as other projects ✅. It also removes the need for the `updategoat` noise 😌.

Finally, this saves me from opening massively embarrassing issues an upstream projects. 😅

## Pull Request Checklist


goat/Task
```diff
@@ -0,0 +1 @@
+../Task
```
JonZeolla, line 1: This is a symlink

ChiefHolland (approved):
LGTM. Love to see unnecessary things go away.

PR #151, opened 2023-08-08
feat: add pre commit validation

ChiefHolland:
# Contributor Comments

This PR adds validation tasks to the test task in order to ensure the pre-commit configuration and hooks files do not have errors and to break the build if they do.

## Pull Request Checklist

Thank you for submitting a contribution to the goat!

In order to streamline the review of your contribution we ask that you review and comply with the below requirements:

- [x] Rebase your branch against the latest commit of the target branch
- [x] If you are adding a dependency, please explain how it was chosen
- [x] If manual testing is needed in order to validate the changes, provide a testing plan and the expected results
- [x] If there is an issue associated with your Pull Request, link the issue to the PR.

PR #153, opened 2023-08-09
fix: pre-commit yaml validation

ChiefHolland:
# Contributor Comments

This PR fixes a bug introduced by the pre-commit validation. If a .pre-commit-*.yaml file doesn't exist, the goat would exit. This was fixed by adding an if statement to the validation bash command.

## Pull Request Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,3 +4,4 @@ repos:
     rev: af8ca5c6840ea54f062a0deb2ff8debec2afcb3f
     hooks:
       - id: seiso-lint
+sdfdfsd
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,4 +4,3 @@ repos:
     rev: af8ca5c6840ea54f062a0deb2ff8debec2afcb3f
     hooks:
       - id: seiso-lint
-sdfdfsd
--- a/Task/Taskfile.yml
+++ b/Task/Taskfile.yml
@@ -68,22 +68,18 @@ tasks:
     dir: ../../..
     cmds:
       - echo "-Validating pre-commit config file"
-      - |
-        if test -f .pre-commit-config.yaml; 
-        then pipenv run pre-commit validate-config .pre-commit-config.yaml; 
-        else echo ".pre-commit-config.yaml does not exist"; 
-        fi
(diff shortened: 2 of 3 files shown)


JonZeolla:
@ChiefHolland would it make sense just to invert the status instead?

https://github.com/SeisoLLC/goat/blob/41b30c54053d7e36def156452d236e7033a9c9f6/Taskfile.yml#L67

ChiefHolland:
On the surface this would seem logical, but in this case I don't think it would work. In the case of .git init, if false, do a thing. In this case, if false, don't do a thing. I couldn't find a way native to Taskfile to skip a task in the case of a false condition. I'll run some tests to verify, though.

JonZeolla (changes requested):
@ChiefHolland I just tested it and the `!` solution works. If you don't disagree, can you use that? It should be cleaner

JonZeolla:
To improve the documentation of the code I referenced would you mind applying this patch?  It's more clear

```bash
diff --git a/Taskfile.yml b/Taskfile.yml
index da3b837..e0919e6 100644
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -62,7 +62,7 @@ tasks:
       for its shared configs, etc.
     internal: true
     status:
-      # Don't update submodules if you aren't in a git repository; quote to avoid yaml intrepretering the ! as a node tag
+      # Only update submodules if you are in a git repository; quote to avoid yaml intrepretering the ! as a node tag
       # https://yaml.org/spec/1.2.2/#691-node-tags
       - '! test -d .git'
     cmds:
@@ -74,7 +74,7 @@ tasks:
     sources:
       - .pre-commit-config.yaml
     status:
-      # Don't install the pre-commit hooks if you aren't in a git repository; quote to avoid yaml intrepretering the ! as a node tag
+      # Only install the pre-commit hooks if you are in a git repository; quote to avoid yaml intrepretering the ! as a node tag
       # https://yaml.org/spec/1.2.2/#691-node-tags
       - '! test -d .git'
     cmds:
```

JonZeolla:
The bug was implemented in 30f9b2273cc05b8a98119244c0f297fdcc84e8bd aka #151

JonZeolla (approved):
LGTM

PR #157, opened 2023-08-11
feat: add sboms and vuln scanning

JonZeolla:
# Contributor Comments

This adds `task`s for generating SBOMs and running vuln scans against those SBOMs.

## Pull Request Checklist

### DIFF
--- a/Task/Taskfile.yml
+++ b/Task/Taskfile.yml
@@ -277,7 +277,7 @@ tasks:
           split: ','
           as: platform
         cmd: |
-          export sanitized_platform=$(echo "{{.platform}}" | sed "s%/%_%g")    \
+          export sanitized_platform=$(echo "{{.platform}}" | sed "s%/%_%g")            \
           && grype sbom:sbom.{{.PROJECT_SLUG}}.{{.VERSION}}.${sanitized_platform}.json \
-               --output json                                                   \
+               --output json                                                           \
                --file vulns.{{.PROJECT_SLUG}}.{{.VERSION}}.${sanitized_platform}.json
--- a/Task/Taskfile.yml
+++ b/Task/Taskfile.yml
@@ -80,7 +80,7 @@ tasks:
       - pipenv run pre-commit validate-manifest .pre-commit-hooks.yaml
     status:
       - '! test -f .pre-commit-hooks.yaml'
- 
+
   build:
     desc: Build the project; docker images, compiled binaries, etc.
     platforms: [linux, darwin]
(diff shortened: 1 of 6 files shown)


Taskfile.yml
```diff
@@ -102,7 +102,7 @@ tasks:
     cmds:
       - task: bash:build
         vars:
-          PLATFORM: '{{if eq .PLATFORM "all"}}{{.SUPPORTED_PLATFORMS}}{{else if .PLATFORM}}{{.PLATFORM}}{{else}}{{.LOCAL_PLATFORM}}{{end}}'
+          PLATFORM: '{{.PLATFORM | default .LOCAL_PLATFORM}}'
```
JonZeolla, line 105: Made the base `Taskfile.yml` a little more simple.  The downside to this approach is that the passing of the `SUPPORTED_PLATFORMS` var is a little less clear because it's in set the including `Taskfile` and only referenced in the included `Taskfile` without being explicitly passed.  But, that's how Taskfiles work /shrug

Task/Taskfile.yml
```diff
@@ -89,7 +89,7 @@ tasks:
       vars: ['VERSION']
     vars:
       VERSION: '{{.VERSION}}'
-      PLATFORM: '{{.PLATFORM | default .LOCAL_PLATFORM}}'
+      PLATFORM: '{{if eq .PLATFORM "all"}}{{.SUPPORTED_PLATFORMS}}{{else if .PLATFORM}}{{.PLATFORM}}{{else}}{{.LOCAL_PLATFORM}}{{end}}'
```
JonZeolla, line 92: This only needs to be done here and not on `publish` because publish will pass along a var that's set, or default to `LOCAL_PLATFORM` if unset like we expect in this logic.  If all is set, it just gets passed and then interpreted / overwritten here

Task/Taskfile.yml
```diff
@@ -239,3 +239,45 @@ tasks:
...
+    desc: Generate project SBOMs
+    dir: ../../..
+    preconditions:
+      - sh: which syft
+        msg: "Syft must be installed and reasonably current"
+    vars:
+      IMAGE_AND_TAG: '{{.IMAGE_NAME}}:{{.VERSION}}'
+      PLATFORM: '{{if eq .PLATFORM "all"}}{{.SUPPORTED_PLATFORMS}}{{else if .PLATFORM}}{{.PLATFORM}}{{else}}{{.LOCAL_PLATFORM}}{{end}}'
+    cmds:
+      - for:
```
JonZeolla, line 255: This is a neat trick, recently introduced to `Taskfile`s and very useful so we don't need to over-manage the syntax we are forced into due to `docker buildx`'s `--platform` comma separated platforms.  We just split on it and make it a local `platform` var, then transform any `/`s in the platform(s) to `_`s for filenames, otherwise the `/` will make folders.

ChiefHolland (approved):
Good stuff!

PR #169, opened 2023-08-31
feat: add task tool install helpers

JonZeolla:
# Contributor Comments

This adds some tool install helpers to make `task init` more powerful in downstream projects, reduce the dev setup steps (no more installing `pipenv`, `syft`, `grype`, etc. manually), and lets us remove a bunch of custom inline logic in the github actions workflow configs.

## Pull Request Checklist

Thank you for submitting a contribution to our project!

- [x] Rebase your branch against the latest commit of the target branch.
- [x] If you are adding a dependency, please explain how it was chosen.
- [x] If manual testing is needed in order to validate the changes, provide a testing plan and the expected results.
- [x] Validate that documentation is accurate and aligned to any project updates or additions.

Don't forget our more detailed contribution guidelines
[here](https://github.com/SeisoLLC/operations/blob/main/documents/software-guidelines.md#contributing-to-a-repository).


Task/Taskfile.yml
```diff
@@ -37,6 +37,55 @@ tasks:
...
+          *${RUNNER_TEMP}/bin*) exit 0 ;;
+          *) echo "${RUNNER_TEMP}/bin" >> "${GITHUB_PATH}" ;;
+        esac
+    requires:
+      vars:
+        - INSTALL_URL
+        - TOOL
+    platforms: [linux]
+    cmds:
+      - curl -sSfL {{.INSTALL_URL}} | sh -s -- -b "${RUNNER_TEMP}/bin"
```
JonZeolla, line 75: This command is likely very specific to the current use case which is the `anchore` install scripts pattern, but we can iterate in the future when it's necessary.

Task/Taskfile.yml
```diff
@@ -37,6 +37,55 @@ tasks:
...
+    internal: true
+    platforms: [darwin]
+    requires:
+      vars: [TOOLS]
+    cmds:
+      - for:
+          var: TOOLS
+          split: ','
+          as: tool
+        cmd: brew install {{.tool}}
```
JonZeolla, line 89: I prefer installing one-at-a-time for troubleshooting and such, but could be convinced to do a `brew install x y z` and cut out the `for`.  `TOOLS` would then become space-separated instead of , separated.

Taskfile.yml
```diff
@@ -19,6 +19,11 @@ includes:
       PYTHON_VERSION: 3.11
       VERSION: '{{.VERSION}}'
 
+  base:
```
JonZeolla, line 22: A divergence from our prior language-specific abstractions; init in the past was entirely independent of the `goat/` but since `init-submodules` happens before `init-install-tools` it shouldn't be a problem.  I think

seanctech (approved):
LGTM 🚀

PR #198, opened 2024-06-04
fix: running the goat locally

JonZeolla:
# Contributor Comments

This fixes a regression from e40e851a4f8925606824d6a6b0b10a9f378ac3ff where it checks for a github action-only variable in a way that fails the overall workload.

## Manual testing

Tested it locally with a `task build` and `task lint` which now works but when testing I had to temporarily remove the following otherwise it would pull from docker hub and overwrite the local build

https://github.com/SeisoLLC/goat/blob/f5005e6636726c17fb3a0c572fd7a608d4c74a80/Task/Taskfile.yml#L99-L102

## Pull Request Checklist

Thank you for submitting a contribution to our project!

- [X] Rebase your branch against the latest commit of the target branch.
- [X] If you are adding a dependency, please explain how it was chosen.
- [X] If manual testing is needed in order to validate the changes, provide a testing plan and the expected results.
- [X] Validate that documentation is accurate and aligned to any project updates or additions.

Don't forget our more detailed contribution guidelines
[here](https://github.com/SeisoLLC/operations/blob/main/documents/software-guidelines.md#contributing-to-a-repository).

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -455,7 +455,7 @@ function rerun_lint() {
 }
 
 function initiate_code_review() {
-  if [ -n "$GITHUB_ACTIONS" ]; then
+  if [[ -n ${GITHUB_ACTIONS:+x} ]]; then
     # Run the Python script
     python /opt/goat/bin/code_review.py
   fi
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,17 @@
 # syntax=docker/dockerfile:1
 
-# TARGETPLATFORM is special cased by docker and doesn't need the inital ARG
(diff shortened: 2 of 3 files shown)


entrypoint.sh
```diff
@@ -455,7 +455,7 @@ function rerun_lint() {
 }
 
 function initiate_code_review() {
-  if [ -n "$GITHUB_ACTIONS" ]; then
+  if [[ -n ${GITHUB_ACTIONS:+x} ]]; then
```
JonZeolla, line 458: This is the primary fix; everything else are updates which `task update` doesn't handle automatically.

ChiefHolland (approved):
LGTM

